PR #75, opened 2021-04-08
Cli guide

wtrocki:
Adding CLI guide to the quickstarts I'm not sure if that is right.

I have noticed that number of quickstarts mix reference to the ascidocs and yml. We can unify that in separate PR.

CC @craicoverflow @DuncanDoyle


bhardesty:
> Was this by BU requirement or simply initial assumption? I have assigned guide that depends on it so wanted to start with refining this one.
I just meant that the content in this guide was originally a series of separate *.adoc files. I combined them into the single README.adoc and edited the content. It could certainly be a quick start, but we need to define the particular use case. It makes sense to me that it would follow the same sequence as the getting started quick start (install > create Kafka instance > create service account > create topic).

PR #80, opened 2021-04-10
Fix quarkus guide link in README

kcbabo:
Note that the service-binding guide doesn't appear to exist anymore. I didn't remove it from the README because I'm not sure if it will be coming back.


wtrocki (approved):
I have PR that adds it but let's remove it for now

PR #81, opened 2021-04-12
Initial import of Quick Starts for DevSandbox.

DuncanDoyle:


wtrocki (approved):
I have been rendering those in our cluster and using them on sandbox beta. All good to merge. UI guide is still in operator - does it make sense to move it here?

DuncanDoyle:
No, that UI guide needs to stay in the Operator. This is the content for Summit. Once we have OpenShift 4.8 and Dev Console support, I want to remove these guides in favour of the ones using the Dev Console feature. These guides really focus on the workshop content (i.e. with that tooling container).

PR #87, opened 2021-04-12
Update links in rhoas cli guide

bhardesty:
Some of the links in the rhoas CLI guide still point to the bf3 repo. I changed them to point to the app-services-cli repo.


bhardesty:
@craicoverflow I assume that we no longer want to point users to the bf3/cli repo to find the rhoas releases. I changed those links to point to the redhat-developer/app-services-guides repo.

wtrocki:
good to have some property in future

craicoverflow:
thank you @bhardesty 🙏🏻, completely overlooked this!

PR #89, opened 2021-04-13
Fix topic creation in Kafka bin scripts quick start

bhardesty:
Previously, the Kafka bin scripts quick start stated that when running kafka-console-producer.sh, the topic would be created automatically. However, this is no longer true as the Kafka instances do not support topic auto-creation. I added a step to create the topic manually using kafka-topics.sh so that messages can be produced and consumed on it.


bhardesty:
@sterobin could you do a quick peer review for this? Also, note that the Kafkacat quick start will need an analogous update.

sterobin (approved):
@bhardesty Looks good. And ack, I'll update accordingly for kafkacat.

PR #90, opened 2021-04-13
Fix attributes and fix topic creation in kafkacat

sterobin:
@bhardesty On the same token, would you mind doing a quick peer review here as well? I also updated the attributes to be more consistent. I did this once before be they keep diverging. And the attributes injection isn't working for me on the new repo, so I just did it manually for now. I'll be asking Pete. I also manually added the signup-link and samples-git-repo variables downstream so that they render in the CP. Thanks.


bhardesty (approved):
This looks good to me. One thing to consider: would it be helpful to add a step to create the topic? If not, your changes read fine as-is.

sterobin:
@bhardesty I wondered the same thing. The only reason I kept it as is, is that the pattern I/we have followed is "enter this command and replace the values accordingly", which in this case includes the topic. I'll think on this though when I do my final UI sanity check and see again in the larger context if it makes sense to call it out more explicitly as a step or if it's a given. Thanks.

PR #92, opened 2021-04-15
Update quick starts per latest UI and dev preview note

sterobin:
@bhardesty , would you mind peer reviewing? I updated against the latest version of the UI and added the dev preview note.


getting-started/README.adoc
```diff
@@ -25,6 +25,11 @@ WARNING: This content is generated by running npm --prefix .build run generate:a
 END GENERATED ATTRIBUTES
 ////
 
+[IMPORTANT]
+====
+{product-long} is currently available for Development Preview. Development Preview releases provide early access to a limited set of features that might not be fully tested and that might change in the final GA version. Users are discouraged from using Development Preview software in production or for business-critical workloads. Limited documentation is available for Development Preview releases and is typically focused on fundamental user goals.
+====
+
```
sterobin, line 32: @bhardesty , @jciroli-rh , @jbyrne-redhat , Paul (don't see his tag), here's the dev preview note. I've added it to all 4 quick starts as the first thing after the title, before the abstract. If Pantheon doesn't do well with it coming before the abstract or if we don't like that, I can move it to after the abstract. It just seemed logical to have it be the very first thing users see. Let me know what you think.

jbyrne-redhat: Looks good, @sterobin.

PaulRMellor: @sterobin (Where's my tag!?). Looks good. :+1:

jciroli-rh: LGTM

sterobin: @PaulRMellor, :) Sorry, your tag isn't showing up for me. I also notice that I don't think you're even listed in the org mgmt as a member of the group, which is probably why. I'll check it out and add you if not. Thanks.

bhardesty: LGTM

bhardesty (approved):

PR #96, opened 2021-04-15
Add note about using the bin scripts on Windows

bhardesty:
The Kafka distribution includes Windows-versions of the bin scripts. I added a note about this so that any potential Windows users are aware.

### DIFF
--- a/kafka-bin-scripts/README.adoc
+++ b/kafka-bin-scripts/README.adoc
@@ -31,6 +31,8 @@ As a developer of applications and services, you can use the Kafka bin scripts t
 The Kafka bin scripts are a set of shell scripts that are included with the Apache Kafka distribution.
 With these bin scripts, you can produce and consume messages for your Kafka instances.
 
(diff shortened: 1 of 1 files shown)


bhardesty:
@sterobin would appreciate a peer review of this. For now, I added the note to the quick start abstract, but not the intro (so it won't be visible in the UI version of the quick start). Not sure if we need it in both places.

wtrocki (approved):

kafka-bin-scripts/README.adoc
```diff
@@ -31,6 +31,8 @@ As a developer of applications and services, you can use the Kafka bin scripts t
 The Kafka bin scripts are a set of shell scripts that are included with the Apache Kafka distribution.
 With these bin scripts, you can produce and consume messages for your Kafka instances.
 
+NOTE: The command examples in this quick start demonstrate how to use the Kafka bin scripts on Linux and macOS. If you're using Windows, you should use the Windows versions of the bin scripts. For example, instead of the `__<Kafka-distribution-dir>__/bin/kafka-console-producer.sh` script, you would use the `__<Kafka-distribution-dir>__\bin\windows\kafka-console-producer.bat` script.
+
```
sterobin, line 35: @bhardesty Trivial, but even with informal tone I think we're still going for minimalism and action-focused language, so I'd still still avoid "should" and "would" unless we have a good reason. I'd just say "If ..., use the Windows ..."  and "script, use the ...". Not critical if you disagree. Just my view.

sterobin (approved):
@bhardesty Done, just a minor comment.

PR #101, opened 2021-04-19
Verify OpenShift guides

wtrocki:
During review I found possible improvements:

- Urls to token page are as plain text. (Adding link doesn't seem to make difference - possibly due to lack of support for links in the list view?
- We request token twice (this will give us 2 different tokens) we could possibly ask user to keep page open and use the same token for both login and connect command. (connect can be run in the interactive mode that will be more impressive - without kafka selection
- quay.io/rhosak/quarkus-kafka-sb-quickstart (we use rhoas and bf2 orgs) I haven't heard about rhosak. How to get access to this org?

### DIFF
--- a/devsandbox-quickstarts/rhosak-devsandbox-connect-cli-toolscontainer-quickstart.yaml
+++ b/devsandbox-quickstarts/rhosak-devsandbox-connect-cli-toolscontainer-quickstart.yaml
@@ -143,7 +143,7 @@ spec:
         #### You have all the necessary information to connect your application to your OpenShift Streams for Apache Kafka instance:
         Have you seen the BootstrapServers field with hostname and port?
 
-        Do you know which authentication mechanism (SASL/PLAIN, SASL/OAUTHBEARER) to use?
+Do you know which authentication mechanism (SASL/PLAIN, SASL/OAUTHBEARER) will be used by Kafka?
 
         Do you know the name of the secret in which your **Client ID** and **Client Secret** are stored?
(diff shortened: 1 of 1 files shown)


devsandbox-quickstarts/rhosak-devsandbox-connect-cli-toolscontainer-quickstart.yaml
```diff
@@ -132,16 +132,16 @@ spec:
...
 
       1. Finally the **KafkaConnection** has a reference to **Service Account Secret** that contains the **Client ID** and **Client Secret** needed to connect to the service. Try to find that configuration in your **KafkaConnection**.
 
     review:
       instructions: |-
         #### You have all the necessary information to connect your application to your OpenShift Streams for Apache Kafka instance:
-        Have you seen the Bootstrap Server hostname and port?
+        Have you seen the BootstrapServers field with hostname and port?
 
         Do you know which authentication mechanism (SASL/PLAIN, SASL/OAUTHBEARER) to use?
```
wtrocki, line 146: I found this sentence confusing. Recomendation would be to remove it as binding is always doing SASL/PLAIN.
My take from this was that as user I could edit those values (where this is not possible :)
```suggestion
```

pmuir (approved):

pmuir:
@DuncanDoyle can you push your quick starts to the rhoas org?

wtrocki:
Actually you are right. I got confused and went to check sbo secret which is different. Closing

PR #104, opened 2021-04-19
Fix Kafka Quarkus Dev QuickStart

b1zzu:
* fix the clone cmd
* fix the path to the sample app

### DIFF
--- a/kafka-quarkus-dev/README.adoc
+++ b/kafka-quarkus-dev/README.adoc
@@ -63,7 +63,7 @@ For this quick start, you'll use the Quarkus sample code from the {product} http
 ----
 $ git clone {samples-git-repo} guides
 ----
-. In your IDE, import the `code-examples/quarkus/kafka-quickstart` directory from the repository that you cloned.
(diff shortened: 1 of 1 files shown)


b1zzu:
@wtrocki I've open this PR against your PR because for example the path is a change that will be introduced from your PR

sterobin (changes requested):
@b1zzu , I see you're making changes to this branch https://github.com/redhat-developer/app-services-guides/tree/dde8e6a8605b879b0126cf052b367d86c5a9a45c  (I assume @wtrocki 's work branch as you mentioned), but it's very different from the main branch https://github.com/redhat-developer/app-services-guides  so the paths to the example code you're modifying in the doc won't be correct in the main branch. Maybe I missed something and this branch is intentionally being kept separate from main?

wtrocki:
@sterobin this was intented as fix to my PR - I'm still working on this

b1zzu:
@sterobin GitHub suggested the reviewers :smile:

PR #107, opened 2021-04-20
add consumer group management commands for rhoas cli

craicoverflow:


wtrocki (approved):
Approved from technical point of view

bhardesty (approved):

PR #110, opened 2021-04-20
fix: Update build trigger of kafka source code to wildcard

wtrocki:
original config did not work. Realized that after merge.
Need another merge to check


wtrocki:
i cannot verify without merge :) so just need approval from anyone

b1zzu (approved):

wtrocki:
Latest image was updated

PR #111, opened 2021-04-21
fix: add route to kuberentes template

wtrocki:
Added route to template.

CC @b1zzu


wtrocki:
No review required - Verified it locally

Issue #112, opened 2021-04-21
kafka-quarkus-quickstart fails if docker daemon is missing or not running


hawkdshrh:
The default build (mvn/mvnw clean install / package) fails if docker is not installed or running on the development host.

wtrocki:
@hawkdshrh Yes. That would be quarkus limitation. I think we might move to JIB builds but when I tried last time we got some issues with those. Looking into this now.

b1zzu:
Hi  @hawkdshrh, this issue should be now fixed on the main branch, can you reverified it

wtrocki:
We have been doing verification for that issue continuously. Closing

PR #114, opened 2021-04-22
Container improvements

wtrocki:
Fixing bug that did build image for local development. I have disabled properties (keeping them there for users to change if needed)


wtrocki:
CC @machi1990 As discussed using Jib for builds. With such simple example, all works fine.

b1zzu:
@wtrocki do you think this cold fix this issue: https://github.com/redhat-developer/app-services-guides/issues/112

wtrocki:
Sorry for not linking issue - this is in fact fix for that issue.

b1zzu (approved):

PR #115, opened 2021-04-22
fix: use latest for cli reference

wtrocki:
Tooling image is probably updated already.


wtrocki:
Closing as @DuncanDoyle would update source. Created issue to deal with the problem of multiple places having those:

PR #119, opened 2021-04-22
Add user restriction note and revert quarkus folder name

sterobin:
@jciroli-rh @sjay72 (your tag isn't here yet), @bhardesty @jbyrne-redhat @PaulRMellor (I actually still don't see your tag), can you review the user restrictions note? I have it at the end of the Kafka instances proc, and I've kept it general so that it can be copied-pasted as needed into other docs. I also excluded details about admin role etc and just focused on what the user sees/knows/cares about. Let me know what you think.


sterobin:
@wtrocki @b1zzu So the name change of the Quarkus qs folder from `quarkus` to `kafka-quarkus-dev` that you introduced in #104 apparently breaks our docs build on the Customer Portal. Pantheon v2 currently has an open bug that prevents folders from being renamed after the initial publish has occurred. It's a terrible problem that will hopefully be resolved fast, but meanwhile, I've had to revert this folder name change to re-enable the Customer Portal build.

So can you pleas review and verify the changes in this PR to that end? I've also added a note in this PR about user restrictions on instances they create and have just added the folder name change to the same PR. Thanks.

getting-started/README.adoc
```diff
@@ -85,6 +85,8 @@ The new Kafka instance is listed in the instances table. When the instance *Stat
 [.screencapture]
 .Kafka instance options menu
 image::sak-kafka-instance-options.png[Image of Kafka instance options menu]
+
+NOTE: Only the user who creates a Kafka instance can edit or delete the instance, access the associated service account and topics, and connect to the instance. All other users can access and connect to only their own instances.
```
jenniferciroli, line 89: Thanks for capturing this! Could we:

- Use second person (you) to match the rest of the text (not sure if that would derail re-use)?
- Drop the second sentence, since it's implied by the first?

jbyrne-redhat, line 89: @sterobin Is it worth also clarifying that you can _see_ instances created by other users in your org?

We say this in the tour:
"Although you can see instances created by other users in your organization, you can't manage or connect to those instances."

We also separately say this about service accounts:
"You can create up to two service accounts that work with your preview instance"

We could possibly have said "work only with _your_ preview instance" in that last statement

A reworked version of your note might read something like:
"Although you can see Kafka instances created by other users in your organization, you can't manage or connect to those instances. Only the user who creates an instance can edit or delete the instance, access the associated service account and topics, or connect to the instance."

sterobin: @jciroli-rh ack. I wanted to do the first but thought I'd keep it general enough so that Paul and John could copy-paste and have it be a general note wherever needed. But what do you think about @jbyrne-redhat 's rewrite below? Maybe continue the convo there.

sterobin: @jbyrne-redhat works for me. @jciroli-rh , wdyt? I know the second sentence still isn't second person, but I figured this note is still general enough for copy-paste purposes as needed and for consistency across multiple docs.

sjay72: What Jen said, although I'm struggling to re-phrase it for second person in my mind. (Opinion: As written doesn't give me heartburn, because I feel like a Note already somewhat breaks the tone; it's naturally different because it's a Note.)

sjay72: I really like that, John.

bhardesty: Works for me too!

wtrocki:
Sounds good. Sorry for troubles.

pmuir (approved):

getting-started/README.adoc
jenniferciroli: love it

jenniferciroli: ack. (back to not being able to resolve)

jenniferciroli (approved):

wtrocki (approved):

PR #121, opened 2021-04-22
Remove note about partitions

sterobin:
Addresses the following Jiras:
* https://issues.redhat.com/browse/MGDSTRM-3039 - Remove note about number of partitions
* https://issues.redhat.com/browse/MGDSTRM-3022 - Clarify Kafkacat with stdin

@b1zzu Can you please review and verify? Fixes your reported bugs.

@b1zzu However, can you please clarify what changes you're suggesting for the `stdin` bit, something about using a `|` instead? I don't follow. Here's how it currently reads:

```
. On the command line, enter the following commands to start Kafkacat in _producer_ mode. This mode enables you to produce messages to your Kafka topic.
+
--
This example uses the SASL/PLAIN authentication mechanism with the server and credential environment variables that you set previously. This example produces messages to a topic in {product} named `my-first-kafka-topic`. Replace the topic name with the relevant topic as needed. The topic that you use in this command must already exist in {product}.

.Starting Kafkacat in producer mode
[source]
----
$ kafkacat -t my-first-kafka-topic -b "$BOOTSTRAP_SERVER" \
 -X security.protocol=SASL_SSL -X sasl.mechanisms=PLAIN \
 -X sasl.username="$USER" \
 -X sasl.password="$PASSWORD" -P
----
```

### DIFF
--- a/devsandbox-quickstarts/rhosak-devsandbox-getting-started-quickstart.yaml
+++ b/devsandbox-quickstarts/rhosak-devsandbox-getting-started-quickstart.yaml
@@ -4,7 +4,7 @@ metadata:
   name: rhosak-devsandbox-getting-started-quickstart
 spec:
   displayName: Getting Started with Red Hat OpenShift Streams for Apache Kafka
-  tags:     
+  tags:
   - streams
   - kafka
   durationMinutes: 10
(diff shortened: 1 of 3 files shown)


wtrocki (approved):

wtrocki:
Requires rebase

b1zzu:
@sterobin In a short section under the kafkacat cmd we could say something like this:

You can also pipe the stdout of another command to kafkacat and send it to the topic like this `echo "foo" | kafkacat [...]`

but I'm also fine with leaving it out if it's going to complicate the much the QS

b1zzu (approved):
There is no point in adding a note for the kafkacat issue because otherwise, we will have to start doing the same for every issue caused by snowflakes user environments

sterobin:
@b1zzu Ack, will just leave out per our convo. Thanks. I've updated in JIRA.

pmuir (approved):

PR #123, opened 2021-04-22
Fixes and improvements from Quarkus qs hackathon.

sterobin:
@bhardesty If you're up for another? Wanted to put this in front of you in case of any impact on other docs besides these few quick starts. Also note the slight change in the Welcome intro text per the hackathon feedback. I agree there too, so made that change in all 4 quick starts. Let me know what you think.

### DIFF
--- a/devsandbox-quickstarts/rhosak-devsandbox-getting-started-quickstart.yaml
+++ b/devsandbox-quickstarts/rhosak-devsandbox-getting-started-quickstart.yaml
@@ -78,7 +78,7 @@ spec:
     description: >-
       After you create a Kafka instance, you can create Kafka topics to start producing and consuming messages in your services.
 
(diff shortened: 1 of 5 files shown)


bhardesty (approved):
Everything looks good to me. At a later date, we may want to circle back to the "click" vs. "select" language once we fully support mobile. I think "click" is totally fine for now, though.

wtrocki (approved):

sterobin:
@bhardesty Yeah, good point. For partly that reason I went with "select", but there was a recurring suggestion in the hackathon that selecting the instance name vs. selecting the instance row wasn't clear enough with how I described it, and that using the different word to emphasize the difference a bit more. I agree to reassess when we go mobile, which will have other implications too. Thanks.

pmuir (approved):

PR #127, opened 2021-04-23
Quickstarts update

wtrocki:
Based of https://github.com/codeready-toolchain/sandbox-sre/tree/master/resources/sandbox.x8i5.p1.openshiftapps.com/configure/member/consolequickstart


wtrocki:
@pmuir this is quick merge - source of the quickstarts (in guides) weren't updated (I have used latest sandbox qucikstarts and updated them in source so we can use them for testing etc.

wtrocki:
Actually they are originaly as templates (not sure why) so we might need to check what was updated in them

pmuir:
Ok, let me know when it needs review

wtrocki:
It is ready now. Sadly files weren't formated using vscode so diff is not useful at all. You all need to believe me that this is exact version that sandbox have

pmuir (approved):

pmuir:
Agreed and done.

wtrocki:
CC @b1zzu to reapply them

devsandbox-quickstarts/rhosak-devsandbox-getting-started-quickstart.yaml
```diff
@@ -5,112 +5,114 @@ metadata:
...
+        1. Click **Create topic** and follow the guided steps to define the topic details. 
+        Click **Next** to complete each step and click **Finish** to complete the setup.
+
+            **Topic name**: Enter a unique topic name, such as ```my-first-kafka-topic```.
+
+            **Partitions**: Set the number of partitions for this topic. This example sets the partition to ```1``` for a single partition. 
+            Partitions are distinct lists of messages within a topic and enable parts of a topic to be distributed over multiple brokers in the cluster. 
+            A topic can contain one or more partitions, enabling producer and consumer loads to be scaled.
+
+            NOTE: You can increase the number of partitions later, but you cannot decrease them.
```
b1zzu, line 92:
```suggestion
```
Not true anymore, is not possible to edit the partitions after creating the topic, it will be enable in the future

wtrocki:
Amazing @b1zzu

wtrocki:
Applything those changes but I guess this is still outdated so another update will be needed.

PR #129, opened 2021-04-26
Remove compact cleanup policy

b1zzu:
Because from the UI my understand is that we preconfigured the **delete** cleanup policy for Topics and is not possible to change I think the quickstart should just tell the user that the policy used is **delete**
image


b1zzu:
@sterobin wdyt? wouldn't make sense to remove the option in this way?

sterobin (commented):
@b1zzu Makes sense if it's true that RHOSAK doesn't support compact and supports only delete. I've pushed a commit to this branch with a few other places where it's mentioned so that this update is complete.

@pmuir Can you confirm if RHOSAK currently supports only delete cleanup policy as suggested in the UI, or if compact is also supported. The UI says only delete, but our docs said both, so need to know whether to update the docs as we have with this PR to be for delete only or leave the docs as is and update the UI to include compact. Or leave the UI and update the docs for now so that they're aligned. Please advise on this docs PR. Thanks.

sterobin:
Confirmed in gchat with @pmuir that the docs are correct and the top-level UI description text could just be clearer.

PR #130, opened 2021-04-26
Fix conditionalization in verification step

bhardesty:
Added "qs" condition to the verification for the first procedure so that it doesn't show up when rendered outside of the UI.


sterobin:
@bhardesty Just following up. Were you holding this for any reason? Just making sure I didn't miss something. Thanks.

wtrocki (approved):
Checked other places as well. This is good to merge

PR #133, opened 2021-04-29
Fix: attributes are not rendered locally

pmuir:


pmuir:
@sterobin this fixes that bug with the attributes not working locally

wtrocki (approved):

sterobin:
@pmuir Excellent. Thank you!

PR #134, opened 2021-04-29
Add JDK prereq to bin scripts.

sterobin:
@pmuir @bhardesty, I got a ping from @jgiardino about the following user feedback for the bin scripts quick start:

>The prerequisites mention only latest JDK for Windows, but I'm on macOS. I'd say we need to add the requirement for the latest JDK regardless of the OS.

He's referring to the Oracle JDK prereq which is mentioned "For Windows". In the Quarkus quick start, we mention both JDK 11+ and Oracle JDK for Windows but in the bin scripts we mention only the Oracle JDK for Windows, nothing about JDK 11+. So I _think_ the user is confused because he's misreading that to mean the JDK 11? Or does the Oracle JDK apply to MacOS too? I've pushed this PR to add the JDK 11 prereq to the bin scripts quick start. Is this the right way to clarify so that it's clear what applies to Windows only, or do we just need to make the Oracle JDK apply to all OS? Not sure what to do here.

Thanks

### DIFF
--- a/getting-started/README.adoc
+++ b/getting-started/README.adoc
@@ -35,11 +35,13 @@ As a developer of applications and services, you can use {product-long} to creat
 
 //For more overview information about {product}, see [variablized link to overview here https://access.redhat.com/documentation/en-us/red_hat_openshift_streams_for_apache_kafka/].
 
-ifndef::community[]
 .Prerequisites
+ifndef::community[]
 * You have a Red Hat account.
 //* You have a subscription to {product-long}. For more information about signing up, see *<@SME: Where to link?>*.
 endif::[]
+* https://adoptopenjdk.net/[JDK^] 11 or later is installed.
+* For Windows, the latest version of https://www.oracle.com/java/technologies/javase-downloads.html[Oracle JDK^] is installed.
 
 // Condition out QS-only content so that it doesn't appear in docs.
 // All QS anchor IDs must be in this alternate anchor ID format `[#anchor-id]` because the ascii splitter relies on the other format `[id="anchor-id"]` to generate module files.
--- a/getting-started/quickstart.yml
+++ b/getting-started/quickstart.yml
@@ -16,6 +16,8 @@ spec:
   prerequisites:
     - A Red Hat identity
     - Access to the Red Hat OpenShift Streams for Apache Kafka environment at https://cloud.redhat.com
+    - JDK 11 or later
+    - For Windows, the latest version of Oracle JDK
   introduction: !snippet README.adoc#introduction
   tasks:
     - !snippet/proc README.adoc#proc-creating-kafka-instance
--- a/kafka-bin-scripts/README.adoc
+++ b/kafka-bin-scripts/README.adoc
@@ -42,6 +42,7 @@ ifndef::community[]
 * You have a Red Hat account.
 endif::[]
 * You have a running Kafka instance in {product}.
+* https://adoptopenjdk.net/[JDK^] 11 or later is installed.
 * For Windows, the latest version of https://www.oracle.com/java/technologies/javase-downloads.html[Oracle JDK^] is installed.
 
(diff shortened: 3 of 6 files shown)


pmuir:
I am really not sure why we would require Oracle JDK - OpenJDK should work. @DuncanDoyle any ideas?

wtrocki:
There was problem with certificate not being approved by valid certification authority in OpenJDK. There were workaround to setup some different certification paths for this on OpenJDK or use Oracle one. 

Oracle one on windows seems to work out of the box so this was one of the suggestion done quickly before summit. 

CC @b1zzu who created original issue

pmuir:
Makes sense - @sterobin how much of that detail should go into the tutorial?

sterobin:
No details imo, just need to know what to tell Mac users, since apparently the Mac user who provided this feedack suggested that it apply to all OS. However, he might just have mistaken the Oracle JDK for JDK 11+ (Open JDK), which does indeed apply to all OS but simply wasn't mentioned in the bin scripts quick start. I guess the questions are: 

* Is JDK 11+ required for bin scripts quick start or only the Quarkus quick start?
* Is the Oracle JDK requirement for bin scripts truly only Windows? How about other Mac and Linux users?

We just need to know which prereqs to put in which quick starts for sure. As far as the why and the details for the quick starts, I don't think that's necessary imo. Deferring to @bhardesty if he disagrees.

pmuir:
JDK 8 will work with the service, but it requires additional configuration, so for the quick starts we should require JDK 11+ wherever we need a JDK. For the service we can require JDK 8+ and provide info on how to configure.

Any JDK will work for Mac and Linux but there are issus with the certificate store in non-Oracle JDKs on Windows, so for the quick start we should require Oracle JDK (and work to fix OpenJDK or and provide info on how to configure). This applies to all quick starts that need a JDK.

b1zzu:
I agree with @pmuir, also maybe should spend some time figuring out how to fix OpenJDK for Windows, and then we can add the instruction to do it instead of requiring the Oracle JDK

sterobin:
@pmuir @b1zzu So if I've understood right, I've added the JDK 11+ and the Oracle JDK (for Windows) prereqs to all four quick starts, to clear up any confusion for users. Is that what you meant and does that address the issue from what you were suggesting?

bhardesty (approved):

getting-started/README.adoc
```diff
@@ -35,11 +35,13 @@ As a developer of applications and services, you can use {product-long} to creat
...
 //For more overview information about {product}, see [variablized link to overview here https://access.redhat.com/documentation/en-us/red_hat_openshift_streams_for_apache_kafka/].
 
-ifndef::community[]
 .Prerequisites
+ifndef::community[]
 * You have a Red Hat account.
 //* You have a subscription to {product-long}. For more information about signing up, see *<@SME: Where to link?>*.
 endif::[]
+* https://adoptopenjdk.net/[JDK^] 11 or later is installed.
+* For Windows, the latest version of https://www.oracle.com/java/technologies/javase-downloads.html[Oracle JDK^] is installed.
```
b1zzu, line 44: Just for confirmation, this section is not rendered in the UI:
image
so this is only for the Documentation right?

sterobin: @b1zzu This exact line, yes, only in docs. But the same prereqs for JDK 11 and Oracle JDK are repeated in the YAML for displaying in the UI. Example, see the yaml file that follows below ...

b1zzu (approved):

wtrocki:
Not sure how this happened but after merging this PR all jobs on master failing because of modular docs splitter issue

wtrocki:
I think that this started to break unrelated to PR. Sorry for confusion

pmuir:
It supposed to check the modular docs build in the PR. Hmm.

PR #137, opened 2021-05-12
Moved the quickstarts to the app-services-operator repository so they …

DuncanDoyle:
…can be automatically installed when the operator is installed on OpenShift/Kubernetes.


wtrocki:
Build is failing because of unrelated reasons. Merging this for clarity

PR #143, opened 2021-05-19
Revert "Add topic configs folder to start"

pmuir:
Reverts redhat-developer/app-services-guides#141


pmuir:
@sterobin fyi this PR caused the build to fail.

PR #144, opened 2021-05-19
Topic configs content

sterobin:
@Simon, ready for your SME review. It's a new doc "Configuring topics in RHOSAK", which still lacks content on consumer groups that's coming soon from @PaulRMellor. For now, it's the sections on reviewing and editing properties and the reference for the properties themselves. We wanted to get your SME review of this portion first, then when the consumer groups content is ready, will ping you for your review of that after.

Use the content preview link below to review, and comment in the topic configs README as needed. You'll see other changed files in this PR that are more tooling and content strategy related, so you can ignore those.

Links:
* [Community content preview](http://file.rdu.redhat.com/~sterobin/MAKDOC-155/)
* [Jira](https://issues.redhat.com/browse/MAKDOC-155)
* [UI fix Jira](https://issues.redhat.com/browse/MGDSTRM-3498) related to this PR


sterobin:
Finished SME review edits from Simon.

@PaulRMellor, ready for your peer review. @b1zzu ready for your QE validation. Be sure to use the content preview link above in the description to review and then add comments as needed to the PR. Thank you both!

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+|Yes
+
+h|Type
+|String
+
+h|Default value
+|None
+
+h|Supported values
+|Letters (Aa-Zz), numbers, underscores ( _ ), or hyphens ( - ), maximum of 249 characters
```
b1zzu, line 113:
```suggestion
|Letters (Aa-Zz), numbers (0-9), underscores ( _ ), or hyphens ( - ), maximum of 249 characters
```
so initially I thought numbers were not included because I just saw the characters within the parentheses, so even if obvious maybe adding the parentheses also for numbers will make it easy to read it

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+[cols="25%,75%"]
+|===
+h|Editable
+|No
+
+h|Type
+|Long
+
+h|Default value
+|`86400000 ms` (1 day)
```
b1zzu, line 398: In some properties I saw the annotation
```
1 week
```
like this, even if the value should be in ms, and this is the only property where the the the value is shown is ms

I think we should unify all of them, I personally find the ms + parentheses annotation more explanatory of the real value used

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+[cols="25%,75%"]
+|===
+h|Editable
+|No
+
+h|Type
+|Long
+
+h|Default value
+|`292271023 years, 2 weeks, 2 days, 13 hours, 12 minutes, 56.064 seconds`
```
b1zzu, line 656: I think the Default value here is just a number representing the number of messages and not a duration

topic-configuration/README.adoc
sterobin: Sure, and that approach is what's shown in the [spec sheet](https://docs.google.com/spreadsheets/d/1HgDxBfy5Zsvoj4Ucv3HCdeifhoBJXpZtsfhTfvuYoPM/edit#gid=0). However, the only reason for the difference was that it was like that in the UI. Meaning, some cases it was shown in ms as shown in the spec, sometimes in the actual value, so I was reflecting what the UI showed. But I agree with you that should be unified, so I've unified all applicable cases to show the ms + parentheses. Done.

topic-configuration/README.adoc
sterobin: @b1zzu Good point actually. But that's what it shows in the UI. See image below. Same default value for both the flush interval message and the flush interval time. So may be a bug?

image

craicoverflow:
```json
{
      "key": "flush.messages",
      "value": "9223372036854775807"
},
{
      "key": "flush.ms",
      "value": "9223372036854775807"
}
```

It appears that the same value is returned from the Kafka Admin API. I'm not sure if that is correct, it could be worth logging an issue on https://github.com/bf2fc6cc711aee1a0c2a/kafka-admin-api

sterobin: @craicoverflow ack. I've created the issue: https://github.com/bf2fc6cc711aee1a0c2a/kafka-admin-api/issues/65

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+ifdef::context[:parent-context: {context}]
+:context: configuring-topics
+
+[IMPORTANT]
+====
+{product-long} is currently available for Development Preview. Development Preview releases provide early access to a limited set of features that might not be fully tested and that might change in the final GA version. Users should not use Development Preview software in production or for business-critical workloads. Limited documentation is available for Development Preview releases and is typically focused on fundamental user goals.
+====
+
+// Purpose statement for the assembly
+[role="_abstract"]
```
PaulRMellor, line 45: Would we want a line of what topics are and how integral they are to using Kafka? And that's why we want to check their properties...

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+As a developer of applications and services, you can refer to the properties of your topics in {product} to better understand how the Kafka implementation is managed for your services. You can edit certain topic properties according to the needs and goals of your services.
```
PaulRMellor, line 46: Does this need to be mentioned? _As a developer of applications and services_. Will this vary according to organisation?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+.Prerequisites
+* You are logged in to the {product} web console.
+* You have created a Kafka instance with at least one Kafka topic in {product}.
+
+.Procedure
+. In the {product} web console, go to *Streams for Apache Kafka* > *Kafka Instances* and click the name of the Kafka instance that contains the topics that you want to configure.
+. In the *Topics* page, click the name of the topic that you want to review or edit.
++
+Alternatively, select the options icon (three vertical dots) for the relevant topic and click *Edit*.
```
PaulRMellor, line 74: Did we settle on `options icon` for the vertical ellipsis/kebab menu? I wonder if there should be a title for the column `Actions` or something.

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+* {base-url}{getting-started-url}[_Getting started with {product}_^]
+* {base-url}{rhoas-cli-url}[_Getting started with the `rhoas` CLI_^]
+* {base-url-cli}{rhoas-cli-ref-url}[_CLI command reference (rhoas)_^]
+
+
+[id="ref-supported-topic-properties_{context}"]
+== Supported topic properties in {product}
+
+[role="_abstract"]
+The following topic properties are supported in {product}. Each listed topic property indicates whether the property is editable or read only, and includes other relevant property attributes for your reference.
```
PaulRMellor, line 89: What do we mean by supported here? Will there be other properties that are set implicitly and managed directly by App Services that we don't see. _Topic properties configured in App Services_?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+=== Core configuration
+
+These properties determine the identity and core behavior of the topic. Before deploying your topic, consider entering all core configuration details for optimal results.
```
PaulRMellor, line 93: _consider_ leaves me wondering whether or not I have to configure all the editable core config.  What are the _optimal results_ we're aiming for?  Maybe the config is important for the performance of the topic and cannot be easily changed once the topic is in use.

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+Name::
++
+--
+The topic name is the unique identifier for the topic within the cluster. You need this to set up your producers and consumers, so make it something memorable.
```
PaulRMellor, line 98: I don't think you can change this later. Should we mention?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+h|Supported values
+|Letters (Aa-Zz), numbers (0-9), underscores ( _ ), or hyphens ( - ), maximum of 249 characters
+|===
+--
+
+Partitions::
++
+--
+Partitions are distinct lists of messages within a topic. Partitions are the main concurrency mechanism in Kafka and enable parts of a topic to be distributed over multiple brokers in the cluster. A topic can contain one or more partitions, enabling producer and consumer loads to be scaled. After you create a topic, you can increase the number of partitions but you cannot decrease it.
```
PaulRMellor, line 120: Wasn't sure what was meant by _concurrency mechanism_ on first reading. I think the sentence still works without that. _Partitions enable parts of a topic to be distributed over multiple brokers in the cluster._

PaulRMellor: Too many _enables_?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+--
+
+=== Messages
+
+These properties control how your messages are handled in the Kafka instance.
+
+Maximum message bytes::
++
+--
+Maximum message bytes is the maximum record batch size.
```
PaulRMellor, line 243: 1 MB is optimal for maximum throughput in most use cases

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+h|Kafka property name
+|`compression.type`
+|===
+--
+
+
+Message format version::
++
+--
+Message format version is the `ApiVersion` value that the broker uses to append messages to topics. This value must be a valid `ApiVersion` value, such as 0.8.2, 0.9.0.0, or 0.10.0.
```
PaulRMellor, line 332: Should we use later versions rather than _0.8.2, 0.9.0.0, or 0.10.0_ as examples?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+[cols="25%,75%"]
+|===
+h|Editable
+|No
+
+h|Type
+|String
+
+h|Default value
+|`2.7-IV2`
```
PaulRMellor, line 344: Will this be 2.8 for next release?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+h|Kafka property name
+|`message.timestamp.difference.max.ms`
+|===
+--
+
+Compression type::
++
+--
+Compression type determines the final compression for the topic. The `Producer` compression type retains the original compression set by the producer.
```
PaulRMellor, line 309: Say anything on the compression types the producer can use?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+h|Kafka property name
+|`cleanup.policy`
+|===
+--
+
+Delete retention time::
++
+--
+Delete retention time is the amount of time that deletion tombstone markers are retained if the topic is compacted.
```
PaulRMellor, line 386:
```suggestion
Delete retention time is the amount of time that deletion tombstone markers are retained if the log is compacted.
```

PaulRMellor: Worth introducing with explanation of tombstones? _Producers send a tombstone message to act as a marker to tell a consumer the value is deleted._

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+h|Kafka property name
+|`num.partitions`
+|===
+--
+
+Replicas::
++
+--
+Replicas are copies of partitions in a topic. Partition replicas are distributed over multiple brokers in the cluster to ensure topic availability if a broker fails. When a follower replica is in sync with a partition leader, the follower replica can become the new partition leader if needed.
```
PaulRMellor, line 145: Replication is **really** important for fault toleration and high availability. Do we want to underline this?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+--
+
+=== Replication
+
+These properties control the behavior of your replicas. Each of these properties impacts every replica created in the topic.
+
+Unclean leader election::
++
+--
+Unclean leader election allows a follower replica that is not in sync with the partition leader to become the leader of the partition. This property provides a way to retain at least partial data if partition leaders are jeopardized. However, this property can lead to data loss, so it is disabled by default.
```
PaulRMellor, line 456: Mention that by disabling this property, clean election is always used? Clean leader election to an in-sync replica guarantees no loss of data.

PaulRMellor: _jeopardized_ might not be clear. Fails or goes offline? Unable to pick up messages?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+[cols="25%,75%"]
+|===
+h|Editable
+|No
+
+h|Type
+|Boolean
+
+h|Default value
+|`Disabled`
```
PaulRMellor, line 468: `false`?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+--
+
+=== Cleanup
+
+These properties control the cleanup processing of the log.
+
+Log segment size::
++
+--
+Log segment size is the size of the log segment files. Log processing actions, such as deletion and compaction, operate on log segments, so a larger setting results in fewer files but less frequent log processing.
```
PaulRMellor, line 482: Logs are a series of segment files...

PaulRMellor: The actions are only performed on older, inactive segments, not the active segment

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+h|Kafka property name
+|`segment.bytes`
+|===
+--
+
+Segment time::
++
+--
+Segment time is the amount of time after which the current log segment is rolled even if the segment file is not full. This property ensures that the topic cleanup policy can delete or compact old messages as needed.
```
PaulRMellor, line 504: Current log segment will be the _active_log segment

PaulRMellor: I don't quite get this sentence: _This property ensures that the topic cleanup policy can delete or compact old messages as needed._ It means that the active segment is rolled every 7 days. Messages older than that are deleted or compacted according to policy.

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+h|Kafka property name
+|`segment.ms`
+|===
+--
+
+Segment jitter time::
++
+--
+Segment jitter time is the maximum delay for log segment rolling. This delay prevents bursts of log segment rolling activity.
```
PaulRMellor, line 526: No delay is set. Should we mention that?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+[cols="25%,75%"]
+|===
+h|Editable
+|No
+
+h|Type
+|Boolean
+
+h|Default value
+|`Disabled`
```
PaulRMellor, line 582: `false`? (Can't see the console. Going by kafka defaults)

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+h|Kafka property name
+|`index.interval.bytes`
+|===
+--
+
+Segment index size::
++
+--
+Segment index size is the size of the index that maps offset to file positions.
```
PaulRMellor, line 618:
```suggestion
Segment index size is the size of the index that maps offsets to file positions.
```

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+--
+
+=== Flush
+
+These properties control the frequency of the flushing of the log.
+
+Flush interval messages::
++
+--
+Flush interval messages is the number of messages between each data flush to the log.
```
PaulRMellor, line 644: I'm not clear on this. Are we flushing from the log cache? So it's how many messages are received before forcing a flush of a topic’s messages to disk?

PaulRMellor: Why's the default in time?

topic-configuration/README.adoc
```diff
@@ -0,0 +1,689 @@
...
+
+h|Kafka property name
+|`flush.messages`
+|===
+--
+
+Flush interval time::
++
+--
+Flush interval time is the amount of time between each data flush to the log.
```
PaulRMellor, line 666: Same as above. It's the interval between flushing of the log cache to disk?

PaulRMellor:
@sterobin Really clear and concise, Stetson.

topic-configuration/README.adoc
sterobin: Yeah, good point. I'm trying to stick to the consistent format of saying the property name as the start of the description. I'm struggling a bit to add some more details without saying "log" so many times. How about this, any better?

>Log segment size is the size of the log segment files **that constitute the log**. Log processing actions, such as deletion and compaction, operate on **old** log segments. A larger setting results in fewer files but less frequent log processing.

topic-configuration/README.adoc
sterobin: Well, it can vary by story. This is just the suggested user-story statement format that I've always been told to user for assembly intros, i.e., "As a ___, I want to ____ so that ____" Makes it immediately clear what user story this assembly addresses, including target user persona, objective, and reason.

topic-configuration/README.adoc
sterobin: Yeah, the official name from the UX team was "menu options icon" or "options icon" for short.

topic-configuration/README.adoc
sterobin: So Kafka offers a ton of [topic properties](https://kafka.apache.org/documentation/#topicconfigs). Our product doesn't support all of them, just supports today what is listed in the UI/CLI/this doc. So by "support" we mean literally that, it supports the following list of available Kafka topic properties. To clarify, I've added "following **Kafka** topic properties. That help?

topic-configuration/README.adoc
sterobin: Yeah, good point. The UI says "we recommend entering ...", but we're trying to avoid "we" and even "recommend" in official docs, and my knee-jerk alternative to "recommend" is to invite users to "consider" doing something. But I think you're right that we should just be more prescriptive. And agreed on optimal, just another alternative I use to let users know that by following the suggestion, they'll be pleased with the results without having to get into details.

I've just simplified as follows. That work for you?

> Before deploying your topic, enter all core configuration properties.

topic-configuration/README.adoc
sterobin: Several properties today can't be changed later, in fact most. But that is evolving and some things are becoming changeable. Since it's not yet determined and a bit of a moving target, I'm hesitant to call it out explicitly unless it's certain, and for now I don't think it's certain.

topic-configuration/README.adoc
sterobin: Yeah, I agree actually, and had removed it from the UI copy when I provided it, but SME reviewers specifically asked to keep it because it's apparently an important phrase that the target user will expect and look for as part of understanding the concept. 

I'm personally okay with having "enables" and then later "enabling"

topic-configuration/README.adoc
sterobin: Ack. I think the idea is that all default values of all properties, especially the properties that users can't change, are set at what we've deemed the optimal setting, given that it's a managed service. So I think I'm okay not to mention it for each property.

topic-configuration/README.adoc
sterobin: Well, it's illustrating the value format, not necessarily the version itself. Note how the second is a 4-digit example.

topic-configuration/README.adoc
sterobin: No idea. Great question.

topic-configuration/README.adoc
sterobin: Not sure I follow. The compression type is set to either `Producer` or `Consumer`, but in our product defaults to only `Producer` and isn't changeable.

topic-configuration/README.adoc
sterobin: Done and done. Thanks!

topic-configuration/README.adoc
sterobin: Sure thing. How about this added line:

>Topic replication is an essential property for fault toleration and high availability.

topic-configuration/README.adoc
sterobin: Yes, already mention that it's disabled by default due to the potential for data loss. And agreed on "jeopardized", changed to "... leaders are lost". That work?

topic-configuration/README.adoc
sterobin: I know, I wish the same. But the actual property value used in the product is `disabled`/`enabled`.

topic-configuration/README.adoc
sterobin: Ditto, from before. Actually `disabled`.

topic-configuration/README.adoc
sterobin: Done

topic-configuration/README.adoc
sterobin: My understanding is that there's a data flush mechanism to the log itself within the topic, and this property dictates how often by count of messages that flush occurs.

And yes, great eye! It's a bug. Davide mentioned that above and I filed an issue for it. Details in this comment above: https://github.com/redhat-developer/app-services-guides/pull/144#discussion_r639548941

topic-configuration/README.adoc
sterobin: Ditto

sterobin:
@PaulRMellor Thank you! I've gone through all and replied, implementing many. Feel free to resolve, respond to, or correct my replies as needed. And I've updated the content preview link in the description above. Same one. If needed.

topic-configuration/README.adoc
PaulRMellor: Looks good

topic-configuration/README.adoc
PaulRMellor: Okay

topic-configuration/README.adoc
PaulRMellor: Helps. Thanks Stetson

topic-configuration/README.adoc
PaulRMellor: Works for me

topic-configuration/README.adoc
PaulRMellor: Makes sense.

topic-configuration/README.adoc
PaulRMellor: Yes. Not clear. I was wondering if we wanted to say anything about supported compression types for the producer (gzip, snappy, zstd)

topic-configuration/README.adoc
PaulRMellor: Like it!

topic-configuration/README.adoc
PaulRMellor: Yes, the segment size doesn't determine the timing of cleanup policy. It was the _ensures_ I wasn't sure about. I think the point being made here is that by rolling the active log we _can_ apply policy on older log segments.

topic-configuration/README.adoc
PaulRMellor: Understood.

PaulRMellor:
@sterobin Thanks for clarifying on the comments, Stetson.

PaulRMellor (approved):

topic-configuration/README.adoc
sterobin: @Simon, what do you think? ^^

topic-configuration/README.adoc
sterobin: Ack, I see. @Simon, what do you think for making this clear/more accurate? ^^

sterobin:
@PaulRMellor Thanks so much! I've just left a couple of comments open for Simon W. to weigh in on when he's back from PTO, since we have some time before end of sprint, and since I'm still waiting to hear back on that flush interval bug meanwhile. Once I get Simon's input, I'll push the last tweaks accordingly and merge. Thanks again!

topic-configuration/README.adoc
sterobin: @b1zzu , @craicoverflow , @PaulRMellor, turns out to be an issue just in the UI, so I've submitted the UI bug here: https://github.com/bf2fc6cc711aee1a0c2a/kafka-ui/issues/128  I closed the old API bug that is now irrelevant. Meanwhile I've updated the docs accordingly. Will resolve this thread once the UI is fixed and I've confirmed for docs. Thanks.

sterobin: Fixed in PR https://github.com/bf2fc6cc711aee1a0c2a/kafka-ui/pull/108. Resolving.

topic-configuration/README.adoc
sjwoodman: I think Paul's suggestion of listing the compression types is a good one

sterobin: Done. Reworded to this:

>Compression type determines the final compression for the topic. The standard compression types are `gzip`, `snappy`, `lz4`, and `zstd`. Additional compression types include `Uncompressed`, which does not compress the topic, and `Producer`, which retains the original compression type set by the producer. By default, the compression type is set to `Producer`.

Good with you Simon?

topic-configuration/README.adoc
sjwoodman: Compaction and deletion are not done on the active log segment AFAIU. This setting says that "even though the `retention.bytes` (for example) hasn't been reached, roll this segment (create a new active segment) so that it can be compacted or deleted". I don't have a problem with the current text but you could change it to `Segment time is the amount of time after which the current log segment is rolled even if the segment file is not full. This enables the segment to be deleted or messages compacted as needed".`

sterobin: @sjwoodman, ah, I see. How about this then. That work?

> Segment time is the amount of time after which the current log segment is rolled even if the segment file is not full. This property enables the segment to be deleted or compacted as needed, even if the log retention limits have not yet been reached.

sjwoodman: LGTM

topic-configuration/README.adoc
sjwoodman: LGTM

PR #207, opened 2021-07-13
Change "xref" to "link" in CLI getting started guide

bhardesty:
This eliminates the use of xrefs, and instead uses the same linking mechanism as the quick starts.


bhardesty:
Verified that each link works.

PR #213, opened 2021-07-20
[DOC] add group ID config to quickstarts

PaulRMellor:
Signed-off-by: prmellor <pmellor@redhat.com>

Updates the quickstarts to include _group ID_ configuration for consumer groups.

### DIFF
--- a/kafka-bin-scripts/README.adoc
+++ b/kafka-bin-scripts/README.adoc
@@ -124,14 +124,13 @@ endif::[]
 
 . Create a file called `{property-file-name}`.
 
-. In the `{property-file-name}` file, add the following content to set the Kafka instance client credentials. Replace the values with your own server and credential information.
+. In the `{property-file-name}` file, set the Kafka instance client credentials and consumer group ID. Replace the values with your own server and credential information.
 +
 --
 ifdef::qs[]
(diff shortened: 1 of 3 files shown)


kafka-bin-scripts/README.adoc
```diff
@@ -131,6 +131,9 @@ ifdef::qs[]
 The `<client_id>` and `<client_secret>` are the generated credentials for your service account. You copied this information previously for the Kafka instance in {product} by selecting the options menu (three vertical dots), clicking *View connection information*, and creating the service account.
 endif::[]
 
+Add a group ID for the consumer to be able to join a consumer group.
```
bhardesty, line 134: It seems a little bit inconsistent to call out the consumer group ID by name, but not the other properties. Maybe it would be better to revise the step to say:

> 3. In the app-services.properties file, set the Kafka instance client credentials and consumer group ID. Replace the values with your own server and credential information.
>
> All consumers with the same group ID belong to the consumer group with that ID.

bhardesty (approved):
Looks great, Paul! I made a suggestion for the bin scripts quick start, but you could also apply it to the others for consistency.

wtrocki (approved):

PaulRMellor:
> Looks great, Paul! I made a suggestion for the bin scripts quick start, but you could also apply it to the others for consistency.

Thanks for checking, @bhardesty . I've updated as suggested

PR #226, opened 2021-07-27
Upgrade quickstarts

jschuler:
To latest `@patternfly/quickstarts` package.


.build/src/app/QuickStartDrawerFederated.tsx
```diff
@@ -1,15 +1,13 @@
+import '@patternfly/react-core/dist/styles/base.css';
```
jschuler, line 1: this will be stripped out in prod

.build/webpack.common.js
```diff
@@ -54,18 +52,6 @@ module.exports = (env, argv) => {
             }
           ]
         },
-        {
```
jschuler, line 57: took it out of common and stuck it in the prod config. this way local dev still shows up fine

.build/package.json
```diff
@@ -55,6 +56,7 @@
     "bundle-profile:analyze": "npm run build:bundle-profile && webpack-bundle-analyzer ./stats.json",
     "start": "webpack-cli serve",
     "start:dev": "webpack serve --hot --color --progress --config webpack.dev.js",
+    "start:prod": "serve dist -l 9001 --cors",
```
jschuler, line 59: to test prod build in the context of the `application-services-ui` container, can first run `npm run build` followed by `npm run start:prod`

pmuir (approved):

.build/webpack.common.js
pmuir: It would be good to try to replicate the prod setup in dev it at all possible i.e. load the pf stylesheet from a cdn.

jschuler: I added a `start:prod` target which will serve up the prod build content so you can load it in `application-services-ui`

pmuir:
@jschuler we have a conflict on this.

wtrocki:
I did run that on localhost for checking what is new. LGTM

dlabaj (approved):
LGTM

dlabaj:
@pmuir Is there anything else blocking this from merging?

.build/webpack.common.js
pmuir: We want to keep the `start:dev` target working for dev - the right way is to just not include the styles in the federated module.

pmuir:
See #244

PR #229, opened 2021-07-29
New quickstart for Quarkus + Kafka + Registry

EricWittmann:
This is a draft proposal for a new quickstart that creates a Quarkus app that produces/consumes Kafka messages and **also** integrates with a Service Registry.  It is based off the existing Quarkus+Kafka quickstart and adds Registry integration.  The quickstart requires the following additional ENV vars to be set:

- `REGISTRY_URL`
- `REGISTRY_IDP_URL`
- `REGISTRY_IDP_REALM`
- `REGISTRY_CLIENT_ID`
- `REGISTRY_CLIENT_SECRET`

TBD: quickstart instructions


wtrocki:
FYI @DuncanDoyle @pmuir 

CC @secondsun We will need to have IDP data in service binding definitions.

code-examples/quarkus-kafka-registry-quickstart/src/main/resources/application.properties
```diff
@@ -0,0 +1,50 @@
...
+# Configure the Kafka sink (we write to it)
+mp.messaging.outgoing.generated-price.connector=smallrye-kafka
+mp.messaging.outgoing.generated-price.topic=prices-complex
+mp.messaging.outgoing.generated-price.value.serializer=io.apicurio.registry.serde.avro.AvroKafkaSerializer
+mp.messaging.outgoing.generated-price.apicurio.registry.url=${REGISTRY_URL}
+mp.messaging.outgoing.generated-price.apicurio.registry.auto-register=true
+mp.messaging.outgoing.generated-price.apicurio.registry.avro-datum-provider=io.apicurio.registry.serde.avro.ReflectAvroDatumProvider
+mp.messaging.outgoing.generated-price.apicurio.auth.service.url=${REGISTRY_IDP_URL}
+mp.messaging.outgoing.generated-price.apicurio.auth.realm=${REGISTRY_IDP_REALM}
+mp.messaging.outgoing.generated-price.apicurio.auth.client.id=${REGISTRY_CLIENT_ID}
```
wtrocki, line 13: Question: Would be able to use the same service account for the service registry and kafka. I see that we have separate CLIENT_ID for kafka and REGISTRY_CLIENT_ID for registry?

EricWittmann: Good question.  Yes we can, and we should.  **However**, right now SR is only in staging, and I was using a production kafka cluster in my testing. So I needed separate SAs...

bibryam: I'd love us to make everything driven by convention for the simpler paths. That is use the same client/secret by default, but with the option to use separate ones if needed. 
Even use default SR and SSO URLs (for production env and the case when there is a single instance) so that the user doesn't have to specify it.

wtrocki:
Will require rebase. I have tried it it from context of registry connection.
This approach will work for service binding and manual env setup.

code-examples/quarkus-kafka-registry-quickstart/src/main/resources/application.properties
pmuir: We have a standard set of env var names for client secret and client id in quick starts - see for example https://github.com/redhat-developer/app-services-guides/tree/main/quarkus#configuring-the-quarkus-example-application-to-connect-to-a-kafka-instance

code-examples/quarkus-kafka-registry-quickstart/README.md
```diff
@@ -0,0 +1,9 @@
+Quarkus Kafka+Registry Quickstart
+=================================
+To see how to use this code example follow the [quick start](../../quarkus).
+
+## Additional information on this code example
+
+This code example is based the [code example from Quarkus](https://github.com/quarkusio/quarkus-quickstarts/blob/main/kafka-quickstart)
```
pmuir, line 7: This URL seems wrong

code-examples/quarkus-kafka-registry-quickstart/README.md
```diff
@@ -0,0 +1,9 @@
+Quarkus Kafka+Registry Quickstart
+=================================
+To see how to use this code example follow the [quick start](../../quarkus).
```
pmuir, line 3: this url seems wrong

code-examples/quarkus-kafka-registry-quickstart/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+## Additional information on this code example
+
+This code example is based the [code example from Quarkus](https://github.com/quarkusio/quarkus-quickstarts/blob/main/kafka-quickstart)
+
+This project illustrates how you can interact with Apache Kafka using MicroProfile Reactive Messaging.
```
pmuir, line 9: This seems to be missing something about service registry

code-examples/quarkus-kafka-registry-quickstart/src/main/resources/application.properties
pmuir: Are we adding a quick start for this? I'm missing the text?

EricWittmann:
@pmuir Sorry, I included "TBD: quickstart instructions" in the description/comment of this PR but that was probably not enough explanation.  But anyway I didn't write any of the instructions yet.  I wanted early feedback to determine if this is the sort of thing we actually wanted.

If it is, I'm happy to update the instructions.

wtrocki:
@carlesarnal What we do with this quickstart? This should be probably closed?

carlesarnal:
> @carlesarnal What we do with this quickstart? This should be probably closed?

This can be closed, yes. The other quickstart replaces this one.

PR #260, opened 2021-08-17
fix: ignore open blocks

pmuir:


wtrocki:
What is the context? Change I should be seeing?

pmuir:
This should allow openblocks to start working in documents.

wtrocki:
Maybe @jschuler can help with review.

wtrocki:
By inspecting and debugging in browser I got it checked

PR #292, opened 2021-09-13
chore: remove invalid env variable

### LINKED ISSUE #291: "Quarkus guide contains Group ID"
Mentions a group id here,  it says consumers with the same id belong in the same group. Couldn't see where I find this id..

wtrocki:
fixes #291 

GROUP_ID env variable is not used in quarkus. It is also not documented causing some issues with completing guide.

### DIFF
--- a/quarkus/README.adoc
+++ b/quarkus/README.adoc
@@ -109,8 +109,6 @@ ifdef::qs[]
 The `<bootstrap_server>` is the bootstrap server endpoint for your service account. The `<oauth_token_endpoint_uri>` is the SASL/OAUTHBEARER token endpoint for the Kafka instance. The `<client_id>` and `<client_secret>` are the generated credentials for your service account. You copied this information previously for the Kafka instance in {product} by selecting the options menu (three vertical dots) and clicking *Connection*.
 endif::[]
 
-All consumers with the same group ID belong to the consumer group with that ID.
-
 .Setting environment variables for server and credentials
 [source,subs="+quotes"]
 ----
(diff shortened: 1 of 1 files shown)


quarkus/README.adoc
```diff
@@ -117,7 +117,6 @@ All consumers with the same group ID belong to the consumer group with that ID.
 $ export BOOTSTRAP_SERVER=__<bootstrap_server>__
 $ export CLIENT_ID=__<client_id>__
 $ export CLIENT_SECRET=__<client_secret>__
-$ export GROUP_ID=__<group_id>__
```
PaulRMellor, line 120: If it doesn't work with Quarkus, are we going to remove the other doc too?
_and consumer group ID as_
_All consumers with the same group ID belong to the consumer group with that ID._

wtrocki: not sure about kafkacat and bin scripts - for those maybe we want to say explicitly what group to use

PaulRMellor: I mean the mentions of consumer groups in the same quarkus quickstart. Above the id you took out.^^

quarkus/README.adoc
```diff
@@ -102,22 +102,19 @@ ifndef::qs[]
 endif::[]
 
 .Procedure
-. On the command line, set the Kafka instance bootstrap server, client credentials, and consumer group ID as environment variables to be used by Quarkus or other applications. Replace the values with your own server and credential information.
+. On the command line, set the Kafka instance bootstrap server, client credentials as environment variables to be used by Quarkus or other applications. Replace the values with your own server and credential information.
```
PaulRMellor, line 105:
```suggestion
. On the command line, set the Kafka instance bootstrap server and client credentials as environment variables to be used by Quarkus or other applications. Replace the values with your own server and credential information.
```

PaulRMellor (approved):
Couple of suggestions

PR #302, opened 2021-09-21
fix: resolve issues with ui-shared

pmuir:
* also upgrade to node 16
* also resolve ws issues in dev-server


.build/package.json
```diff
@@ -67,7 +67,7 @@
     "generate:modular-docs": "npm-run-all --parallel generate:modular-docs-sources fetch:templates --sequential generate:modular-docs-output"
   },
   "dependencies": {
-    "@rhoas/app-services-ui-shared": "^0.12.1",
+    "@rhoas/app-services-ui-shared": "^0.12.2",
```
wtrocki, line 70:
```suggestion
    "@rhoas/app-services-ui-shared": "0.12.2",
```

PR #324, opened 2021-09-30
Service registry RHOAS CLI

PaulRMellor:
**Updates to the Getting Started CLI guides**

- Creates three guides as follows:
-- Installing and configuring the rhoas CLI
-- Getting started with the rhoas CLI for OpenShift Streams for Apache Kafka
-- Getting started with the rhoas CLI for OpenShift Service Registry
- Removes install steps from the _getting tarted_ guides, replaced by prereq that goes to the new common _Install_guide
- Updates links and attributes to reflect the new structure
- Removes the _Commands for..._ tables to replace with `-h` list and link to command reference guide
- Pushes the _Using_ procedures for Kafka Getting Started up a level (==) and removes the discrete title flag

PR #325, opened 2021-09-30
New access mgmt doc, updated getting started doc, and related access mgmt updates

sterobin:
Jira: https://issues.redhat.com/browse/MGDSTRM-5414

Summary of updates:
* New access mgmt doc
* New section in getting started quick start about service account access
* New prerequisite in other quick starts where needed for adding service account permissions
* Related updates throughout for access mgmt

Doc previews (VPN required):
* [Managing account access in RHOSAK](http://file.rdu.redhat.com/~sterobin/MGDSTRM-5414_access-mgmt/README.html)
* [Getting started > Setting permissions for a service account](http://file.rdu.redhat.com/~sterobin/MGDSTRM-5414_getting-started/README.html#proc-setting-service-account-permissions_getting-started)

**NOTE about ACL terminology:** Kafka, Confluence, etc, use the term "ACLs" both for the actual ACL (the table/list) and for individual ACL entries, which is highly confusing. In reality, the whole table/list is the ACL, where individual entries are "ACL entries", "permission entries", "permission settings", or just "permissions". I've tried hard to not deviate from Kafka, but it just is terrible to refer to individual permission entries as "ACLs". So I've not done that for now.


access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+
+[id="con-access-management_{context}"]
+== Access management in {product}
+
+[role="_abstract"]
+{product} uses Access Control Lists (ACLs) provided by Kafka that enable an organization administrator to manage how user accounts and service accounts are permitted to interact with Kafka resources, such as instances and topics. A user account enables users in your organization to access resources. A service account enables your application or tool to connect securely to your resources.
+
+Administrators can allow or deny operations that specific accounts or that all accounts in the organization can perform for Kafka resources. A resource in an ACL can be a Kafka instance, topic, consumer group, or producer transaction. Administrators can also allow or deny access for other users to define account permissions for an instance.
+
+Users who aren't administrators can fully access a Kafka instance that they create and can manage the level of access for service accounts that they create for their own instances. However, only administrators can manage the level of access of other users or of service accounts created by other users for a Kafka instance.
```
MikeEdgar, line 67: Same comment as above. An instance owner has full control to manage access for any user or service account in the organization.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+
+=== Supported ACL permissions in {product}
+
+An ACL acts as a mapping of permitted operations on specified resources for a selected account or for all accounts in an organization. An account can be a user account or a service account. Operations correspond to one or more Kafka APIs or request types that are applicable for the specified resource.
+
+For example, a `Read` operation for a `Topic` resource corresponds to the `Fetch`, `OffsetCommit`, and `TxnOffsetCommit` Kafka requests. A `Write` operation for a `Topic` resource corresponds to the `Produce` and `AddPartitionsToTxn` Kafka requests.
+
+The following table lists the supported ACL permissions in {product}.
+
+NOTE: The resource identifiers `Starts with` and `Is` support the wildcard character `*` to denote any occurrences of the specified resource. For example, `Topic` = `is *` means any topics in a Kafka instance.
```
MikeEdgar, line 89: The wildcard is only valid for `Is` and not `Starts with`.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+
+[role="_additional-resources"]
+.Additional resources
+https://kafka.apache.org/documentation/#security_authz_primitives[Authorization Primitives^] in Kafka documentation
+
+[id="proc-setting-permissions_{context}"]
+== Setting account permissions in {product}
+
+[role="_abstract"]
+In {product}, you can create Access Control Lists (ACLs) in Kafka instances and set permissions for how user accounts or service accounts can interact with an instance and its resources. Only administrators can manage the level of access of other users or of service accounts created by other users for a Kafka instance. If you're managing only service accounts that you created for a Kafka instance that you created, you don't need administrator rights.
```
MikeEdgar, line 168: Same as other comment regarding an instance owner's ability to set ACLs for any account.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+* The user account or service account that you're setting permissions for has been created in the organization.
+
+.Procedure
+. In the {product} web console, go to *Streams for Apache Kafka* > *Kafka Instances* and click the name of the Kafka instance that you want to set permissions for.
+. Click the *Permissions* tab to view the current ACL permissions for this instance.
++
+--
+By default, new Kafka instances contain the following permissions. These permissions allow all accounts in the organization to view topics in the instance, but not to produce or consume messages.
+
+NOTE: The wildcard character `*` denotes any occurrences of the specified resource. For example, `Topic` = `is *` means any topics in a Kafka instance.
```
MikeEdgar, line 182:
```suggestion
NOTE: The wildcard character `*` denotes any occurrences of the specified resource. For example, `Topic` = `is *` means any topic in a Kafka instance.
```

access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+
+.Default ACL permissions for new Kafka instances
+[cols="25%,25%,25%,25%"]
+|===
+h|Account
+h|Resource
+h|Access type
+h|Operation
```
MikeEdgar, line 190: This table should also include `Describe` access on the cluster (limited to the describe ACLs API)

access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+|===
+h|Resource type
+h|Resource identifier
+h|Access type
+h|Operation
+
+|`Topic`
+|`Is` = `*`
+|`Allow`
+|`Write`
```
MikeEdgar, line 241: `Write` should be `All` to enable full produce/consume access.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+h|Resource type
+h|Resource identifier
+h|Access type
+h|Operation
+
+|`srvc-acct-1a2b3c4d-...`
+|`Topic`
+|`Is` = `*`
+|`Allow`
+|`Write`
```
MikeEdgar, line 295: Use `All` for full producer/consumer.

getting-started/README.adoc
```diff
@@ -170,6 +171,58 @@ ifndef::qs[]
...
+|===
+h|Resource type
+h|Resource identifier
+h|Access type
+h|Operation
+
+|`Topic`
+|`Is` = `*`
+|`Allow`
+|`Write`
```
MikeEdgar, line 203:
```suggestion
|`All`
```

MikeEdgar (commented):
@sterobin - a few comments/suggestions. Are the larger HTML versions just generated from the adocs?

access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+====
+
+// Purpose statement for the assembly
+[role="_abstract"]
+As an organization administrator in {product}, you can manage the level of access for user accounts and service accounts in Kafka instances. You can allow or deny operations that specific accounts or that all accounts in your organization can perform for an instance or for certain resources in that instance, such as topics. You can also allow or deny access for other users or service accounts to define account permissions for an instance.
+
+If you're not an administrator in {product}, your own user account has full access to the Kafka instances that you create, and you can manage the level of access for service accounts that you create for your own instances. However, you can't manage the level of access of other users or of service accounts created by other users for your Kafka instances or for any other Kafka instances.
```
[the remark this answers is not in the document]
sterobin: Per discussion in chat, changed to the following. Will likewise update throughout where applicable. I didn't realize that this story was geared to instance owners and not admins. I got it backward.

>As an owner of a Kafka instance in {product}, you can manage the level of access for other user accounts and service accounts in that Kafka instance that you created. You can allow or deny operations that specific accounts or that all accounts in your organization can perform for your instance or for certain resources in that instance, such as topics. You can also allow other users or service accounts to manage access for you in your instance.

access-mgmt/README.adoc
sterobin: Fixed similar to above, to this:

>{product} uses Access Control Lists (ACLs) provided by Kafka that enable you to manage how other user accounts and service accounts are permitted to interact with the Kafka resources that you create. A resource in an ACL can be a Kafka instance, topic, consumer group, or producer transaction. A user account enables users in your organization to access your resources. A service account enables your application or tool to connect securely to your instance and access your resources.

access-mgmt/README.adoc
sterobin: Yeah, I had done this intentionally throughout (any topics, any consumer groups, etc.) to emphasize plural, and had it as "all" at first, but I think the wildcard is more correctly "any", and you're right that grammatically it's "any topic", "any consumer group", etc. Will correct throughout.

access-mgmt/README.adoc
sterobin: Done.

access-mgmt/README.adoc
sterobin: Fixed. My bad, and that seems more logical. Thanks.

getting-started/README.adoc
sterobin: Done

access-mgmt/README.adoc
sterobin: Done.

access-mgmt/README.adoc
sterobin: Done

access-mgmt/README.adoc
sterobin: Done

sterobin:
@MikeEdgar I implemented your feedback, resolved several comments, and left just the ones to get your final okay on. I also updated the doc preview links above. Refresh your browser when you open them to get the latest.

Regarding the HTML files, I generate those locally to verify output, and then use those to provide the doc previews. I meant to delete them from the repo and have done so now. I need to create a gitignore file but have just been lazy. Will get around to that.

access-mgmt/README.adoc
MikeEdgar: Yes, that looks good.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,478 @@
...
+(Any consumer group)
+|`Allow`
+|`Describe`
+
+|`All accounts`
+|`Transactional ID is *`
+
+(Any producer)
+|`Allow`
+|`Describe`
```
MikeEdgar, line 210: Missed this earlier, but I don't think we need to have describe on transactions enabled by default. There is no use for it in the UI that I am aware of.

sterobin: Sure, what should the default setting for transactions be then? And will the change to the UI be made for the milestone? This table simply describes what the default ACL is in the UI, so I have to follow whatever will actually be true in the UI for the milestone.

MikeEdgar: I have no configuration at all for transactions in my notes. The ones out there currently (the compatibility rules) do still have "all" operations allowed for any transactional ID. We might need confirmation on this one.

sterobin: @MikeEdgar Okay, I'll wait to hear back from you on that. Feel free to ping others here if needed. Thanks.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+
+[role="_additional-resources"]
+.Additional resources
+* link:https://kafka.apache.org/documentation/#security_authz[Authorization and ACLs^] in Kafka documentation
+
+=== Supported ACL permissions in {product}
+
+An ACL acts as a mapping of permitted operations on specified resources for a selected account or for all accounts in an organization. An account can be a user account or a service account. Operations correspond to one or more Kafka APIs or request types that are applicable for the specified resource.
+
+For example, a `Read` operation for a `Topic` resource corresponds to the `Fetch`, `OffsetCommit`, and `TxnOffsetCommit` Kafka requests. A `Write` operation for a `Topic` resource corresponds to the `Produce` and `AddPartitionsToTxn` Kafka requests.
```
DuncanDoyle, line 85: This is almost a verbatim copy of the text from Confluent, without a reference/citation to the original text: https://docs.confluent.io/platform/current/kafka/authorization.html

I'm not sure what the license is on those docs, and therefore I don't know if we can simply copy their text, but just wanted to point this out to avoid potential problems.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,473 @@
...
+.. Click *Next*.
+.. Under *Review existing permissions*, verify the current permission settings for this instance. You can delete existing permissions now if needed, or you can wait to delete existing permissions later from the main *Permissions* page.
++
+If you previously selected a specific account, you can delete only permission entries that apply to individual accounts. If you previously selected *All accounts*, you can delete only permission entries that apply to all accounts.
+
+.. Under *Assign Permissions*, set the permissions for the specified account or all accounts for a resource type, such as a topic, and click *Add* to continue adding permissions for other resources as needed.
++
+--
+For example, when you create a new service account, add the following permissions in the Kafka instance that you want the account to access. These permissions enable applications associated with the service account to produce and consume messages in any topics in the instance, using any consumer group and any producer.
```
DuncanDoyle: This reads like I "HAVE" to do that for each service account, i.e. that I have to add those exact permissions. However, I could also have a service account that is only allowed to consume from a given instance, but not produce to it. So I'd rather say something like:

"For example, when you create a new service account, you can add the the following permissions in the Kafka instance that you want the account to access."

sterobin: Sure, I just thought we wanted to be more prescriptive with users here, to tell them to just create these permissions for new service accounts. But you're right, it's one thing to be prescriptive in the getting started like we've done to achieve a specific tutorial outcome, but here it's just an option. So I've tweaked as you said, with "you can" and "in this example" to emphasize that it's just one option.

access-mgmt/README.adoc
sterobin: @DuncanDoyle Good point. I wasn't actually trying to randomly copy Confluent, they just have exactly what I was going for, so I did copy, but just as a basis. I wanted to point out that each operation corresponds to Kafka APIs or request types that relate to the specified resource, so that users have a bit more context about the operations and how they relate back to Kafka. I also wanted to give a couple of examples and read and write are the most common so I wanted those.

Anyway, I've re-modified the previous paragraph a bit more to distinguish. As per the examples, I believe it's different enough to be okay, imo. I don't know what more to change than I already have, and I want to make the examples clear. And I don't want to refer to Confluent because I'm already referring to Kafka, which is the true source.

access-mgmt/README.adoc
MikeEdgar: @sterobin - the defaults for transactional IDs can be removed.

sterobin: @MikeEdgar Awesome, fixed. Resolving. Thank you!

getting-started/README.adoc
```diff
@@ -170,6 +171,60 @@ ifndef::qs[]
...
+h|Resource identifier
+h|Access type
+h|Operation
+
+|`Topic`
+|`Is` = `*`
+|`Allow`
+|`All`
+
+|`Consumer group`
```
wtrocki, line 207:
```suggestion
|`Group`
```

wtrocki: If we want to be aligned with how actual values are represented. 
If we focus on what we present in the UI we might likely get disconnected with CLI/Operator etc.

MikeEdgar (approved):
Looks good from the perspective of ACL defaults and how the back-end supports managing the rules presently.

getting-started/README.adoc
```diff
@@ -170,6 +171,60 @@ ifndef::qs[]
 . Verify that your service account was successfully created in the *Streams for Apache Kafka* > *Service Accounts* page.
 endif::[]
 
+[id="proc-setting-service-account-permissions_{context}"]
+== Setting permissions for a service account in a Kafka instance in {product}
+
+{product} uses Access Control Lists (ACLs) provided by Kafka that enable you to manage how other user accounts and service accounts are permitted to interact with the Kafka resources that you create. When you create a service account to connect to a Kafka instance, you must also set the appropriate level of permissions for that new account in the ACL of the instance.
```
jciroli-rh, line 177: (1) 
> that enable you to manage how other user accounts and service accounts are permitted to interact with the Kafka resources that you create.

I stumbled over this a bit -- does it still make sense if it's shortened to something like one of the following?

- ... manage how other users and service accounts can interact with ...
- ... manage access to the Kafka resources you create

(2) Can we switch from passive without anthropomorphizing? "Kafka includes Access Control Lists (ACLs) so you can manage ..."
 
(3) Maybe flipping the order of this paragraph and/or reconsidering some of the content might help address our often-lengthy intros and create a closer connection with what the user just did? 

- The service account you just created can only describe/view your Kafka instance and any other resources  (maybe too technical to say "describe") 
- You need to use the Access Control Lists (ACLs) Kafka provides to give the service account permission to publish, consume, etc. 
- Maybe this is where we send them off to the more detailed info, esp. re: user accounts, etc.? I will defer to you, the SMEs, and QE here. :D

getting-started/README.adoc
```diff
@@ -170,6 +171,60 @@ ifndef::qs[]
...
+
+The permissions described in this procedure enable applications associated with the service account to fully interact with the resources in the instance.
```
jciroli-rh, line 179: Would this work in a bit more informal style? It feels kind of stuffy for a quick start in the UI. :D 

Let's give your new service account the permissions it needs to allow associated applications to fully interact with the Kafka instance and resources 

(Okay, maybe that's a touch too informal, but ... )

getting-started/README.adoc
```diff
@@ -170,6 +171,60 @@ ifndef::qs[]
...
+
+.Procedure
+. In the *Kafka Instances* page of the web console, click the name of the Kafka instance that you want the service account to access.
+. Click the *Permissions* tab to view the current ACL for this instance.
+. Click *Manage access*, use the *Account* drop-down menu to select the service account that you previously created, and click *Next*.
+. Under *Assign Permissions*, use the drop-down menus to set the following permissions for the specified resource types for this service account. Click *Add* to add each new resource permission.
++
+--
+These permissions enable applications associated with the service account to produce and consume messages in any topic in the instance, using any consumer group and any producer.
+
```
jciroli-rh, line 193: AH, see, this is the info that it would have been good to have at the beginning, where I noted it felt a little stuffy.

kafka-bin-scripts/README.adoc
```diff
@@ -94,6 +95,7 @@ To enable the Kafka scripts to access a Kafka instance, you must configure the c
 .Prerequisites
 ifndef::qs[]
 * You have the generated credentials for your service account. You copied this information previously for the Kafka instance in {product} by selecting the options menu (three vertical dots), clicking *Connection*, and creating the service account.
```
jciroli-rh, line 97: I've always wondered -- are we assuming that the user who is following non-Getting-Started quick starts is the same user who created the instance -- and thus the service account and its generated credentials? That's how it reads, so I'm assuming yes?

kafka-bin-scripts/README.adoc
```diff
@@ -94,6 +95,7 @@ To enable the Kafka scripts to access a Kafka instance, you must configure the c
...
+* You've set the permissions for your service account to access the Kafka instance resources. You used the *Permissions* tab of the Kafka instance to set these permissions after you created the service account.
```
jciroli-rh, line 98: I'm wondering if there's a good way to streamline this and make it clear that both of these prereqs are related to the service account that you created for the Kafka instance ...

- Are we assuming that the user has used the UI to create the service account and set permissions?
- Are we assuming that the same user who created the instance and service accounts is using this (or any other post-Getting Started quick start)?

If yes to both, maybe it becomes something like:
- You've created a service account and assigned it permissions to access the Kafka instance resources. 
   - For help creating service accounts and assigning permissions, see [link]
- You gave that service account permissions to access the Kafka instance resources

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+
+[id="con-access-management_{context}"]
+== Access management in {product}
+
+[role="_abstract"]
+{product} uses Access Control Lists (ACLs) provided by Kafka that enable you to manage how other user accounts and service accounts are permitted to interact with the Kafka resources that you create. You can manage access for only the Kafka instances that you create or for instances that the owner has enabled you to access and alter.
+
+A resource in an ACL can be a Kafka instance, topic, consumer group, or producer transaction. A user account enables users in your organization to access your resources. A service account enables your application or tool to connect securely to your instance and access your resources.
+
+You can define ACLs for the following components:
```
jgiardino, line 67: This statement sounds off to me. The components you describe are combined to create a single ACL, i.e. these are components or properties _of_ an ACL.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+
+[IMPORTANT]
+====
+{product-long} is currently available for Development Preview. Development Preview releases provide early access to a limited set of features that might not be fully tested and that might change in the final GA version. Users should not use Development Preview software in production or for business-critical workloads. Limited documentation is available for Development Preview releases and is typically focused on fundamental user goals.
+====
+
+// Purpose statement for the assembly
+[role="_abstract"]
+As an owner of a Kafka instance in {product}, you can manage the level of access that other user accounts and service accounts have to your instance. You can allow or deny operations that specific accounts or that all accounts in your organization can perform for your instance or for certain resources in that instance, such as topics. You can also allow other users or service accounts to manage access to the resources in your instance for you.
+
```
jciroli-rh, line 54: I'm tripping on the second sentence -- maybe the length or the order? I wish this wasn't an abstract, because a bulleted list would be awesome. 

Also, there's a bit of overlap between this abstract and the next one -- is there something that could be omitted from this one or vice versa?

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+//Additional line break to resolve mod docs generation error, not sure why. Leaving for now. (Stetson, 20 May 2021)
+
+[id="con-access-management_{context}"]
+== Access management in {product}
+
+[role="_abstract"]
+{product} uses Access Control Lists (ACLs) provided by Kafka that enable you to manage how other user accounts and service accounts are permitted to interact with the Kafka resources that you create. You can manage access for only the Kafka instances that you create or for instances that the owner has enabled you to access and alter.
+
+A resource in an ACL can be a Kafka instance, topic, consumer group, or producer transaction. A user account enables users in your organization to access your resources. A service account enables your application or tool to connect securely to your instance and access your resources.
+
```
jciroli-rh, line 66: I still feel a little confused about user account vs. service account, and I'm not sure I am following how it's related to the sentence above. :(

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+You can define ACLs for the following components:
+
+* A single account or all accounts within the organization
```
jciroli-rh, line 69: Does this mean a user account, a service account, or both?

wtrocki:
Would we have some ACL documentation that explains how to handle some cases and what verbs can be used?

This doc does good job at explaining how to use UI.
Would we have another for how to secure kafka for common use cases?

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+
+If two or more permission settings in an ACL match a request being made to the Kafka broker and at least one of the matching permissions specifies that the action is denied, then the request is denied.
+
+[role="_additional-resources"]
+.Additional resources
+* link:https://kafka.apache.org/documentation/#security_authz[Authorization and ACLs^] in Kafka documentation
+
+=== Supported ACL permissions in {product}
+
+An ACL acts as a mapping of permitted operations on specified resources for a selected account or for all accounts in an organization. An account can be a user account or a service account. Operations correspond to Kafka APIs or request types that relate to the specified resource.
```
jciroli-rh, line 83: Ah, here is a little more context about what "account" means. Might be useful to put this up a bit higher and clarify before users get down here?

getting-started/README.adoc
```diff
@@ -170,6 +171,60 @@ ifndef::qs[]
...
+* You've created a service account that you want to allow to interact with the running Kafka instance.
+
+.Procedure
+. In the *Kafka Instances* page of the web console, click the name of the Kafka instance that you want the service account to access.
+. Click the *Permissions* tab to view the current ACL for this instance.
+. Click *Manage access*, use the *Account* drop-down menu to select the service account that you previously created, and click *Next*.
+. Under *Assign Permissions*, use the drop-down menus to set the following permissions for the specified resource types for this service account. Click *Add* to add each new resource permission.
++
+--
+These permissions enable applications associated with the service account to produce and consume messages in any topic in the instance, using any consumer group and any producer.
```
wtrocki, line 192: This also enables sa to create and delete topics

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+
+[role="_additional-resources"]
+.Additional resources
+https://kafka.apache.org/documentation/#security_authz_primitives[Authorization Primitives^] in Kafka documentation
+
+[id="proc-setting-permissions_{context}"]
+== Setting account permissions in a Kafka instance in {product}
+
+[role="_abstract"]
+In {product}, you can create Access Control Lists (ACLs) in your Kafka instances and set permissions for how other user accounts or service accounts can interact with an instance and its resources. You can manage access for only the Kafka instances that you create or for the instances that the owner has enabled you to access and alter.
```
jciroli-rh, line 168: For searchability, could we reverse a bit to "Create Access Control Lists (ACLs) in your OpenShift for Apache Kafka instances to set permissions..." 
or
"Use Access Control Lists (ACLs) to manage access to Kafka instances and resources in [long honkin' product name]"

getting-started/README.adoc
```diff
@@ -170,6 +171,60 @@ ifndef::qs[]
...
+.ACL permissions for a new service account
+[cols="25%,25%,25%,25%"]
+|===
+h|Resource type
+h|Resource identifier
+h|Access type
+h|Operation
+
+|`Topic`
+|`Is` = `*`
```
wtrocki, line 203: I would rather see examples that focus on enabling topics with prefix and only produce and consume.

I know that @DuncanDoyle was asking us to not open kafkas to world in operator

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+|===
+--
+. Use the *Permissions* page to set permissions for a new account, add permissions to an existing account, or delete account permissions in this instance.
++
+To set permissions for a new account in this instance, follow these steps:
+
+.. Click *Manage access*.
+.. In the *Account* drop-down menu, select the new user account or service account that you want to set permissions for. You can also select *All accounts* to set permissions for all user accounts and service accounts in the organization.
+.. Click *Next*.
+.. Under *Review existing permissions*, verify the current permission settings for this instance. You can delete existing permissions now if needed, or you can wait to delete existing permissions later from the main *Permissions* page.
```
jgiardino, line 220: Maybe something like this instead?
> ...verify the current permission settings _for the selected account_ for this instance...

This section doesn't show all permissions for the instance, it only shows the permissions that affect the selected account.

getting-started/README.adoc
wtrocki, line 207: Confluent using group and we are using consumer group. As person that done kafka acl before this raises question;

Are those guys have custom ACL objects or use confluent?

Since we do not list all possible values, objects this will be hard to answer by reading this guide

wtrocki: Adding (Group) at the end will solve it IMHO

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+.Prerequisites
+* You've created a Kafka instance and the instance is in *Ready* state.
+* The user account or service account that you're setting permissions for has been created in the organization.
+
+.Procedure
+. In the {product} web console, go to *Streams for Apache Kafka* > *Kafka Instances* and click the name of the Kafka instance that you want to set permissions for.
+. Click the *Permissions* tab to view the current ACL permissions for this instance.
++
+--
+By default, new Kafka instances contain the following permissions. These permissions allow all accounts in the organization to view the instance permissions and to view topics in the instance, but not to produce or consume messages.
```
jciroli-rh, line 179: This feels like it belongs in the earlier concept topic so it isn't making the reader scroll so much to the next step. Maybe move the details out, and keep the general note/reminder that all new instances have default permissions that allow view, but not produce and consume. (You could also put the brief reminder before the procedure, too - can't remember if the template allows for it after the abstract or prereqs).

access-mgmt/README.adoc
jciroli-rh, line 220: page, or tab? Have seen it referred to both ways and want to be sure we're following our own style guide on UI item labels. :)

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+
+|`Transactional ID`
+|`Is` = `*`
+|`Allow`
+|`All`
+|===
+--
+.. Click *Save* to finish.
+
+To add permissions to an existing account in this instance, follow these steps:
```
jciroli-rh, line 255: This gets a bit lost at the end of the current procedure. Worth pulling it out on its own (so, setting permissions for a new instance vs for an existing one)?

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+.. For the account that you want to add permissions to, select the options icon (three vertical dots) for that entry and click *Manage*.
+.. Under *Review existing permissions*, verify the current permission settings for this instance. You can delete existing permissions now if needed, or you can wait to delete existing permissions later from the main *Permissions* page.
++
+If you selected a permission entry that applies to a specific account, you can delete only permission entries that apply to individual accounts. If you selected a permission entry that applies to all accounts, you can delete only permission entries that apply to all accounts.
+
+.. Under *Assign Permissions*, set the additional permissions for that account or all accounts for a resource type, such as a topic, and click *Add* to continue adding permissions for other resources as needed.
+.. Click *Save* to finish.
+
+To delete existing account permissions in this instance, use the following options:
+
```
jciroli-rh, line 266: Again, this feels like we have a lot of procedures in a single module. (Or maybe I'm looking at this incorrectly?)

getting-started/README.adoc
```diff
@@ -170,6 +171,60 @@ ifndef::qs[]
...
+
+The permissions described in this procedure enable applications associated with the service account to fully interact with the resources in the instance.
+
+.Prerequisites
+* You've created a Kafka instance and the instance is in *Ready* state.
+* You've created a service account that you want to allow to interact with the running Kafka instance.
+
+.Procedure
+. In the *Kafka Instances* page of the web console, click the name of the Kafka instance that you want the service account to access.
+. Click the *Permissions* tab to view the current ACL for this instance.
```
jgiardino, line 187: We are changing the tab label from *Permissions* to *Access*.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,469 @@
...
+[role="_abstract"]
+In {product}, you can create Access Control Lists (ACLs) in your Kafka instances and set permissions for how other user accounts or service accounts can interact with an instance and its resources. You can manage access for only the Kafka instances that you create or for the instances that the owner has enabled you to access and alter.
+
+.Prerequisites
+* You've created a Kafka instance and the instance is in *Ready* state.
+* The user account or service account that you're setting permissions for has been created in the organization.
+
+.Procedure
+. In the {product} web console, go to *Streams for Apache Kafka* > *Kafka Instances* and click the name of the Kafka instance that you want to set permissions for.
+. Click the *Permissions* tab to view the current ACL permissions for this instance.
```
jgiardino, line 176: We are changing the tab label from *Permissions* to *Access*.

jciroli-rh:
Reviewed. LMK if there is something you need me to do differently - I started a review, but after leaving comments, it still asks me to start a review. Ugh.

kafka-bin-scripts/README.adoc
```diff
@@ -135,6 +137,7 @@ You can use the `kafka-console-producer` script to produce messages to Kafka top
 * You have a running Kafka instance in {product}.
 ifndef::qs[]
 * You have the bootstrap server endpoint for your service account. You copied this information previously for the Kafka instance in {product} by selecting the options menu (three vertical dots) and clicking *Connection*.
+* You've set the permissions for your service account to access the Kafka instance resources. You used the *Permissions* tab of the Kafka instance to set these permissions after you created the service account.
```
jgiardino, line 140: We are changing the tab label from *Permissions* to *Access*.

quarkus/README.adoc
```diff
@@ -102,6 +103,7 @@ Quarkus applications use https://github.com/eclipse/microprofile-reactive-messag
 .Prerequisites
 ifndef::qs[]
 * You have the bootstrap server endpoint, the service account credentials, and the SASL/OAUTHBEARER token endpoint for the Kafka instance. You copied this information previously for the Kafka instance in {product} by selecting the options menu (three vertical dots) and clicking *Connection*.
+* You've set the permissions for your service account to access the Kafka instance resources. You used the *Permissions* tab of the Kafka instance to set these permissions after you created the service account.
```
jgiardino, line 106: We are changing the tab label from *Permissions* to *Access*.

kafka-bin-scripts/README.adoc
jgiardino, line 98: We are changing the tab label from *Permissions* to *Access*.

kafkacat/README.adoc
```diff
@@ -103,6 +104,7 @@ NOTE: Kafkacat does not yet fully support SASL/OAUTHBEARER authentication, so co
 .Prerequisites
 ifndef::qs[]
 * You have the bootstrap server endpoint and the generated credentials for your service account. You copied this information previously for the Kafka instance in {product} by selecting the options menu (three vertical dots) and clicking *Connection*.
+* You've set the permissions for your service account to access the Kafka instance resources. You used the *Permissions* tab of the Kafka instance to set these permissions after you created the service account.
```
jgiardino, line 107: We are changing the tab label from *Permissions* to *Access*.

jciroli-rh (commented):
Left comments - some may be refinements to make as we go.

wtrocki:
Really good job! Left some comments.

I think the main problem with ACL is that doc in upstream reference is very hard to digest. Kafka provides pluggable mechanism for ACL and upstream docs cover every possible case

https://kafka.apache.org/documentation/#security_authz

The most clean and simple doc for ACL is on that page:

https://docs.confluent.io/platform/current/kafka/authorization.html

kafka-bin-scripts/README.adoc
jgiardino: Separate but maybe related comment to @jciroli-rh's 🤔 

This statement seems to imply that a service account that you create from the drawer is somehow tied to a Kafka instance. That wasn't really the case at summit. But going forward, there's only a relationship between the two through the process of defining access on the Kafka instance. 
> You copied this information previously for the Kafka instance

To @jciroli-rh's comment, there are at least 2 places, soon to be 3, where you can create a service account:
* the Service Accounts page
* the Connection drawer for a Kafka instance
* eventually the Connection drawer for a Service Registry instance

getting-started/README.adoc
wtrocki, line 177: Will add link to doc here to make in clear. See comment below listing our options

kafka-bin-scripts/README.adoc
```diff
@@ -135,6 +137,7 @@ You can use the `kafka-console-producer` script to produce messages to Kafka top
 * You have a running Kafka instance in {product}.
 ifndef::qs[]
 * You have the bootstrap server endpoint for your service account. You copied this information previously for the Kafka instance in {product} by selecting the options menu (three vertical dots) and clicking *Connection*.
```
jgiardino, line 139: Is it more accurate to say this?:
> You have the bootstrap server endpoint for your Kafka instance.

getting-started/README.adoc
wtrocki, line 179: I will specifically add that we doing this in UI

getting-started/README.adoc
```diff
@@ -170,6 +171,60 @@ ifndef::qs[]
...
+[id="proc-setting-service-account-permissions_{context}"]
+== Setting permissions for a service account in a Kafka instance in {product}
+
+{product} uses Access Control Lists (ACLs) provided by Kafka that enable you to manage how other user accounts and service accounts are permitted to interact with the Kafka resources that you create. When you create a service account to connect to a Kafka instance, you must also set the appropriate level of permissions for that new account in the ACL of the instance.
+
+The permissions described in this procedure enable applications associated with the service account to fully interact with the resources in the instance.
+
+.Prerequisites
+* You've created a Kafka instance and the instance is in *Ready* state.
+* You've created a service account that you want to allow to interact with the running Kafka instance.
```
wtrocki, line 183: created service account doesnt interact with kafka. It is used as credentials when making api call.

Best to say service account that will be used when connecting to Kafka from your app or service

jgiardino (commented):
These contents look really good! I made a few comments based on a few things that seemed off, and also to let you know that the tab label is going to change from "Permissions" to "Access" in this PR: https://github.com/bf2fc6cc711aee1a0c2a/kafka-ui/pull/385/files

kafka-bin-scripts/README.adoc
wtrocki, line 98: So we do not recomend or going to test bin scripts acl cli? I find it odd to tell person to go to UI

access-mgmt/README.adoc
jgiardino: Doh, I missed this in my earlier comments because I was searching for "tab".

We are changing the label *Permissions* to *Access*. 

We also refer to this as a _tab_ in the UI, but in this case I'm wondering if _main page_ makes more sense. When we say _Access tab_, the user's current context is somewhere else besides that tab. But if you are already in that tab, and open the modal, then how do you refer to the contents _behind_ the modal. I actually don't mind the use of _main page_ in this case to refer to the contents behind the modal.

kafka-bin-scripts/README.adoc
wtrocki: Same would apply to rhoas cli. If we want to get acl enabled before we add support we need the same note in rhoas cli tutorial
(after merge of #324 )

kafka-bin-scripts/README.adoc
wtrocki: @jgiardino I was about to comment that

getting-started/README.adoc
```diff
@@ -170,6 +171,60 @@ ifndef::qs[]
...
+.Procedure
+. In the *Kafka Instances* page of the web console, click the name of the Kafka instance that you want the service account to access.
+. Click the *Permissions* tab to view the current ACL for this instance.
+. Click *Manage access*, use the *Account* drop-down menu to select the service account that you previously created, and click *Next*.
+. Under *Assign Permissions*, use the drop-down menus to set the following permissions for the specified resource types for this service account. Click *Add* to add each new resource permission.
++
+--
+These permissions enable applications associated with the service account to produce and consume messages in any topic in the instance, using any consumer group and any producer.
+
+.ACL permissions for a new service account
```
wtrocki, line 194: I would add `Example.. `

getting-started/README.adoc
sterobin: (1) Yeah, I started much more concisely, but then wondered which accounts exactly, so I said "user accounts and service acounts". Then by "can manage" to we mean are able to or are permitted to. So every time I streamlined it, I felt that it introduced some doubt, and since this concept of acess mgmt here is so prone to ambiguities, I opted for the lengthier but clearer explanation to try to avoid the doubt.

(2) I'd have to think on that. At first, I don't know that it feels right.

(3) Great idea here. I had kind of retained the structure similar to the other proc in access mgmt doc, but you're right because out of the context of the access mgmt, the focus should be the action as it falls in this sequence.

I've changed the first paragraph to this. Better?

>After you create a service account to connect to a Kafka instance, you must also set the appropriate level of access for that new account in the Access Control List (ACL) of the Kafka instance. {product} uses ACLs provided by Kafka that enable you to manage how other user accounts and service accounts are permitted to interact with the Kafka resources that you create.

getting-started/README.adoc
sterobin: Well, as you pointed out in your comment later, I mention it here so that the user knows exactly what the permissions in this example will enable the service account to do. But on second thought, I'd actually rather just delete this sentence and leave what I have later. Users can use this for whatever level of access is appropriate, and the example I've given later is just that. Done.

getting-started/README.adoc
sterobin: Yeah, previous reply.

kafka-bin-scripts/README.adoc
sterobin: @jciroli-rh and @jgiardino You're both right here. We've had to minimize cross-refs in our quick starts in the UI and isolate cross-refs instead in "Additional resources" as much as possible, because we don't yet have a good way of cross-reffing in-UI. Especially since prereqs should and will be shown in UI eventually. This is highly impractical most of the time, but one way we've gotten around it is by at least jogging the user's memory like this. But again, you're correct that the direction is getting outdated with new UI features. Also not always the case that the user created these things from the UI to begin with, so better to poise it as just an option or instruction to do, regardless of how they did it the first time.

So how about this rewrite for both prereqs. It's more detailed than we typically go with prereqs, but to be super clear on what to do, and to avoid sending them on a link chase (and we're asked to avoid linking anyway), this is a compromise. Wdyt?

> * You have the generated credentials for your service account. You copied this information previously for the Kafka instance in {product}. To regenerate the credentials, open the Application Services web console and use the *Service Accounts* page to find your service account and update the credentials.
> * You've set the permissions for your service account to access the Kafka instance resources. To verify the current permissions, open the {product} web console, select your Kafka instance, and use the *Access* tab of the Kafka instance to find your service account permission settings.

Then for the quick start procedures that require just bootstrap server endpoint, like later in bin scripts and in Kcat:

>* You have the bootstrap server endpoint for your service account. You copied this information previously for the Kafka instance in {product}. To relocate the server endpoint, open the {product} web console, select your Kafka instance, select the options menu (three vertical dots), and click *Connection*.
>* You've set the permissions for your service account to access the Kafka instance resources. To verify the current permissions, open the {product} web console, select your Kafka instance, and use the *Access* tab of the Kafka instance to find your service account permission settings.


And then for quick start procedures that require multiple or all of the above, like Quarkus:

>* You have the bootstrap server endpoint and the generated credentials for your service account. You copied this information previously for the Kafka instance in {product}. 
To relocate the server endpoint, open the {product} web console, select your Kafka instance, select the options menu (three vertical dots), and click *Connection*.
To regenerate the credentials, open the Application Services web console and use the *Service Accounts* page to find your service account and update the credentials.

>* You've set the permissions for your service account to access the Kafka instance resources. To verify the current permissions, open the {product} web console, select your Kafka instance, and use the *Access* tab of the Kafka instance to find your service account permission settings.

kafka-bin-scripts/README.adoc
sterobin: See my previous comment. We're limited by the fact that we can't use cross-refs in quick start content that appears in the UI, and prereqs will be in the UI soon.

access-mgmt/README.adoc
sterobin: Yep, good catch. Changed to "ACL permissions", which is one way around that issue. Been fighting a battle on not calling individual ACL permissions/entries just "ACLs" themselves, which is how others and even Kafka often refer to them.

access-mgmt/README.adoc
sterobin: I know, it has really been something to try to explain things clearly but concisely. I can't do a ton about the overlap because the purpose of the abstract is to essentially give a digest of it, and so will inherently repeat key things, but I've tried to streamline a bit more.

So how about this, better?

>As an owner of a Kafka instance in {product}, you can manage the level of access that other user accounts and service accounts have to your instance. You can allow or deny access to your instance for specific accounts or for all accounts in your organization. You can also allow other users or service accounts to manage the level of access to your instance for you.

>You can manage access for only the Kafka instances that you create or for instances that the owner has enabled you to access and alter.

access-mgmt/README.adoc
sterobin: Sure, I also saw your comment later that you thought would be good to bring up.

So how about this, better?

>{product} uses Access Control Lists (ACLs) provided by Kafka that enable you to manage how other user accounts and service accounts are permitted to interact with the Kafka resources that you create. You can manage access for only the Kafka instances that you create or for instances that the owner has enabled you to access and alter.

>An account in {product} is either a user account or a service account. A user account enables users in your organization to access your resources. A service account enables your application or tool to connect securely to your instance and access your resources.

>A resource in an ACL can be a Kafka instance, topic, consumer group, or producer transaction. You use the ACL to define how specific accounts or all accounts in an organization are permitted to access these resources.

access-mgmt/README.adoc
sterobin: Clarified in my update above, so I think we're good. Wdyt?

access-mgmt/README.adoc
sterobin: Yep, see above.

access-mgmt/README.adoc
sterobin: I actually thought that providing the product as context was best for searchability, plus it makes it less clunky, and plus we seem to have a pattern of that around. As I try some of these other alternatives, it just doesn't flow as well to my ear and removes the important context for when users hit it from a search. Not sure how to alter it and still achieve that, and still not have it clunky.

access-mgmt/README.adoc
sterobin: @jgiardino I'm not sure what you mean. As a user, I can _view_ all the permission settings regardless of what account I'm using, BUT I can only _update/delete_ other permission settings that are the same as what I'm adding, whether a single account or all accounts.

Can you clarify what you mean by "affect the selected account"? What permissions then wouldn't be shown?

access-mgmt/README.adoc
sterobin: You're right! I had it here initially so newcomers would have an idea what to expect when they get there, but I agree. I've moved that bit up above in the supported ACL section right after the list of supported options, and it flows much better to say what's supported and then what's default there. And I just removed this altogether here from the proc, not needed imo. Thanks!

access-mgmt/README.adoc
sterobin: Right, this has come up a few times. The tab is what you click, open, etc., the page is what you view, go to, navigate to, etc. Just as in real life: You thumb to the tab to view a page. So as @jgiardino described.

So in sum, it's "Click the *Access* tab" but "In the *Access* page, ..." Same with *Connection* in the drawer: "Click the *Connection* tab, ...", but "In the *Connection* page, ..." etc.

access-mgmt/README.adoc
sterobin: I went around the block on this: So normally, I'd actually just make the procedure about adding the permissions for a new account, and mention in passing along the way that to hit up an existing one, use the options menu, and to delete, likewise. This is how we've done it with creating a service account, where we say how to do it anew, and have a last step just saying, to update, rename, delete, whatever, go to the Service Accounts page.

However, because managing access tends to be very confusing about what's what, I decided to be a little less concise (again) to make it more laid out. The reason I don't want to go with individual procs is that it's unmerited imo, and unecessarily splits things up into what would then be too narrow of procedures.

So my compromise is not the full streamlined way (all together with brief mentions) and not the full un-streamlined way (individual procs), but a middle-road option of zeroing in on the core task that just has some options. This is also a common approach in the IBM style guide, to make a step of the overall step, with unnumbered listed options and their sub-steps if any.

However, what I should've done and didn't, especially since it's more than two options, is bullet them. I've done that now and it really helps. Wdyt?

image

image

access-mgmt/README.adoc
sterobin: Ditto

getting-started/README.adoc
sterobin: Ack. I've updated throughout.

kafka-bin-scripts/README.adoc
sterobin: Great catch! Fixed in other fixes above, and throughout docs. Thanks!

getting-started/README.adoc
sterobin: Good call! I've changed to this here and all other places where applicable:

That work?

>the permissions enable applications associated with the service account to create and delete topics in the instance, to produce and consume messages in any topic in the instance, and to use any consumer group and any producer.

getting-started/README.adoc
sterobin: @wtrocki Sure. That's a question for the DXP and UX team though. Today the product uses `Consumer group`, so the docs need to document what's in the product.

getting-started/README.adoc
sterobin: @wtrocki Agreed, already did so in the access mgmt procedure and other sections there, pointing to Kafka for more details. I've just added a Kafka link here as well. Thanks.

getting-started/README.adoc
sterobin: No longer applicable per changes above. Thanks.

getting-started/README.adoc
sterobin: Oops, yep. I went around the block on that wording and missed that.

How about this? This is more consistent with what we say elsewhere and more accurate. Work for you?

>You've created a service account that you want to allow to access the running Kafka instance.

kafka-bin-scripts/README.adoc
sterobin: Agreed. For now I've reworded in the fixes above to just suggest they use the UI for most ease. They can certainly use the CLI, but we've not yet documented ACLs in the CLI and it's not clear if it will make the external milestone.

getting-started/README.adoc
sterobin: Sure. I intentionally didn't say "Example" here but did say that in the access mgmt doc, because here we're prescriptively telling the user to just do this literally, not do something like this, where it's just an example. But I do see the benefit of labeling as such for consistency. Thanks.

sterobin:
@jgiardino @jciroli-rh I finished applying your feedback. Have a look back through, resolve all comments that you're good with, and re-reply where needed. I've updated the doc preview links above, but **refresh the browser when you load the preview to get the latest**.

@wtrocki I also applied what I could from your feedback, thanks. Have a look through, resolve what you're good with, and re-reply where needed. Regarding other ACL scenarios as you mentioned, what's in the docs today is what was determined and suggested by dev in this [common use cases](https://docs.google.com/document/d/1wywQUcxfQVP80bxbkxo4BfDR2Hvut67hQENd7_ExAiY/edit) doc. So it would be a larger conversation I think.

DuncanDoyle (approved):

wtrocki (approved):

access-mgmt/README.adoc
jgiardino: 👍 Thanks for the clarification! This distinction makes sense for the Access tab/page. 
The Connection example seems a little odd to refer to it as a page, but it's not like I have a better suggestion. 🤷

access-mgmt/README.adoc
```diff
@@ -0,0 +1,471 @@
...
+== Access management in {product}
+
+[role="_abstract"]
+{product} uses Access Control Lists (ACLs) provided by Kafka that enable you to manage how other user accounts and service accounts are permitted to interact with the Kafka resources that you create. You can manage access for only the Kafka instances that you create or for instances that the owner has enabled you to access and alter.
+
+An account in {product} is either a user account or a service account. A user account enables users in your organization to access your resources. A service account enables your application or tool to connect securely to your instance and access your resources.
+
+A resource in an ACL can be a Kafka instance, topic, consumer group, or producer transaction. You use the ACL to define how specific accounts or all accounts in an organization are permitted to access these resources.
+
+You can define ACL permissions for the following components:
```
jgiardino, line 69: This statement still seems off. These components you list are what make up one ACL. So something like this sounds more accurate:

> You can define ACL permissions using the following components:

or

> ACL permissions consist of the following properties: 

But I also am really unfamiliar with how to talk about ACLs, so take this with a grain of salt. I tried to find something comparable, and here's the [example from Confluent](https://docs.confluent.io/platform/current/kafka/authorization.html#acl-format), but it's not an exact match for how we're describing it here.

access-mgmt/README.adoc
jgiardino: A few lines up, you have this step:
> In the *Account* drop-down menu, select the new user account or service account that you want to set permissions for.

After they complete this step, the permissions that display in the *Review existing permissions* are only the permissions that affect the account selected in that step. 

So if there are 3 ACLs defined for the Kafka instance, one for _All accounts_, one for _sterobin_, one for _jgiardino_, then this is what you'd see:
* If you selected _jgiardino_ in the *Account* dropdown, then you would see the ACLs for _All accounts_ and _jgiardino_, but not the one for _sterobin_
* If you selected _All accounts_ in the *Account* dropdown, then you would see the ACL for _All accounts_, but not the ACLs for _sterobin_ or  _jgiardino_

getting-started/README.adoc
jgiardino: Here is [documentation from Confluent describing what a group](https://docs.confluent.io/platform/current/kafka/authorization.html#resources) is, but I'm not really sure how their description relates to our service.

We are currently using "Consumer group" in the UI, and I don't remember this point about "group" coming up in any discussions. This might be something to revisit with SMEs.

kafka-bin-scripts/README.adoc
sterobin: Ugh, yeah, sorry. It's correct in the docs, just old copy-paste. Fixed now in my previous comment.

access-mgmt/README.adoc
sterobin: Ah, I got it now. Sorry, I didn't realize that, but I see that now.

I've reworded then as follows. Again a tad wordy, but when I make it more concise, it's just not quite clear. This to me is clearest, though not as concise. Wdyt?

>The *Review existing permissions* section lists any permission settings for this instance that apply to all accounts in the organization and any permission settings that apply to the same account that you previously selected, if applicable. You can delete existing permissions now if needed, or you can wait to delete existing permissions later from the main *Access* page.

>If you previously selected a specific account, you can delete only permission entries that apply to individual accounts. If you previously selected *All accounts*, you can delete only permission entries that apply to all accounts.

access-mgmt/README.adoc
sterobin: Oh, I get it! Sorry, no idea why I wasn't getting that, and you're exactly right.

I've reworded as follows. Better?

>An ACL permission setting typically consists of the following components:

* A single named account or all accounts within the organization that you want to manage access for
* A single named resource, all resources of a particular type (such as a topic, consumer group, or transactional ID), or all resources of a particular type with a specified prefix
* A single operation (such as `Write`) or all operations for the specified resource or resources

access-mgmt/README.adoc
jciroli-rh: Ok -- if you have the chance, maybe check with Ben? If not, we can revisit later if needed. :)

access-mgmt/README.adoc
jciroli-rh: Stetson, if you have what you need here, I'm okay with resolving this - didn't want to do so in case you still needed this info.

kafka-bin-scripts/README.adoc
jciroli-rh: The context in this thread has been super helpful -- thank you both, Stetson and Jenn. If you have what you need here, we can resolve the comment.

access-mgmt/README.adoc
sterobin: @bhardesty Would you mind weighing in on this question? ^^ Would love your perspective.

jciroli-rh (approved):
I think these changes look great - nice job!

I left a few comments unresolved in case you still needed the info (e.g., there were comments from Jenn or others in there). Feel free to resolve (or make me do it!) if they're done and dusted. :D

access-mgmt/README.adoc
bhardesty: I like the original title the best. Setting permissions is the goal, ACLs are an implementation detail. Also, the intro para explains the use of ACLs, so we don't completely lose searchability.

bhardesty: Regarding the intro paragraph, I think it makes sense to first convey the context, and then describe what you can do within that context. So I think I like the original formulation the best, unless it proves to create significant searchability issues.

access-mgmt/README.adoc
```diff
@@ -0,0 +1,477 @@
...
+
+=== Supported ACL permissions in {product}
+
+An ACL acts as a mapping of permitted operations on specified resources for a selected account or for all accounts in an organization. An account can be either a user account or a service account. Operations correspond to Kafka APIs or request types that relate to the specified resource.
+
+For example, a `Read` operation for a `Topic` resource corresponds to the `Fetch`, `OffsetCommit`, and `TxnOffsetCommit` Kafka requests. A `Write` operation for a `Topic` resource corresponds to the `Produce` and `AddPartitionsToTxn` Kafka requests.
+
+The following table lists the supported ACL permissions in {product}.
+
+NOTE: The resource identifier `Is` supports the wildcard character `*` to denote any occurrences of the specified resource. For example, `Topic` = `is *` means any topic in a Kafka instance.
```
jbyrne-redhat, line 93: Should the second occurrence of `Is` be capitalized, too, for consistency?

jbyrne-redhat: I might be harping on this now, but I also ended up playing a game of spot the difference between table 2 and table 3. Is there a difference between `Resource` (table 2) and `Resource type` + `Resource identifier` (table 3)?

access-mgmt/README.adoc
```diff
@@ -0,0 +1,477 @@
...
+|`Topic`
+|`Is` = `*`
+|`Allow`
+|`All`
+
+|`All accounts`
+|`Consumer group`
+|`Is` = `*`
+|`Allow`
+|`All`
```
jbyrne-redhat, line 342: This is likely some ignorance on my part, but I noticed that the operation for a _service account_ to use any consumer group is `Read`, while for _human user accounts_ and `All accounts` it is `All`. I presume there's a good reason for this, and it's not a mistake? :)

getting-started/README.adoc
```diff
@@ -160,16 +161,72 @@ NOTE: SASL/PLAIN authentication is also available for tools and libraries that d
...
+
+.Prerequisites
+* You've created a Kafka instance and the instance is in *Ready* state.
+* You've created a service account that you want to allow to access the running Kafka instance.
+
+.Procedure
+. In the *Kafka Instances* page of the web console, click the name of the Kafka instance that you want the service account to access.
+. Click the *Access* tab to view the current ACL for this instance.
+. Click *Manage access*, use the *Account* drop-down menu to select the service account that you previously created, and click *Next*.
+. Under *Assign Permissions*, use the drop-down menus to set the following permissions for the specified resource types for this service account. Click *Add* to add each new resource permission.
```
jbyrne-redhat, line 189: Per an earlier comment, I suggest "use the drop-down menus to set the permissions **shown in the table**..."

jbyrne-redhat (approved):
Looks excellent, @sterobin. Just a few minor nit picks and questions. Great job.

PR #333, opened 2021-10-08
Add placeholder values registry quickstart

carlesarnal:
This PR fixes the auth values for now until the fix is released on Quarkus *and* also removes the old quickstart, which for some reason was not deleted before.


smccarthy-ie:
Excellent, thanks @carlesarnal I've removed setting the `OAUTH_SERVER_URL` and `OAUTH_REALM` env vars from the instructions :)

PR #354, opened 2021-10-19
Fix Kafka topic create options for Kafka 3

b1zzu:
I tried the quickstart using Kafka 3.0.0 scripts which is the last version that can be downloaded and to create a topic the `--partitions` and `--replication-factor` options are mandatory


sterobin:
Thanks @tombentley !

@b1zzu I pinged around and found that Duncan was on PTO, Wojciech sick, and Pete referred me to Tom, who has approved. Merging.

PR #358, opened 2021-10-21
fix: improve the image handling

pmuir:
* don’t load images that we don’t need
* fix the path for images we do need


wtrocki:
Verified locally. Merging now!

PR #361, opened 2021-10-27
Fix: update to newer splitter

pmuir:


wtrocki:
@pmuir I do not think we need splitter in the repo.
We should do something like :

```
wget https://github.com/LightGuard/asciidoc-splitter/releases/download/v1.5.0/asciidoc-splitter-1.5.0-jar-with-dependencies.jar
``` 

to download it. Binary in the repo will make it very large to be consumed by users wanting to use quickstarts with slower internet connectivity. (Splitting guides is probably good idea)

wtrocki:
Closing. I will remove file from the repo and use wget to get file locally.

PR #365, opened 2021-11-05
use update ACL table format

craicoverflow:
Update to ACL table format to coincide with https://github.com/redhat-developer/app-services-cli/pull/1292


wtrocki:
Changes were released. Merging

PR #368, opened 2021-11-23
fix: use asciidoc splitter from maven repo rather than binary in repo

pmuir:


.build/package.json
```diff
@@ -56,15 +56,16 @@
...
     "clean": "rimraf dist tmp && gh-pages-clean",
-    "generate:attributes": "node ./utils/attributes.js",
-    "generate:modular-docs-sources": "node ./utils/modular-docs.js",
     "commitandpush:modular-docs": "gh-pages -d ./tmp/post-splitter -b modular-docs",
     "publish:modular-docs": "npm run generate:modular-docs-sources && npm run commitandpush:modular-docs",
     "fetch:templates": "node ./utils/templates.js",
+    "fetch:splitter": "mvn -f utils/pom.xml package",
     "build:generated-assemblies": "asciidoctor --failure-level=WARN  --template-dir=templates --destination-dir=tmp/generated/assemblies tmp/post-splitter/assemblies/assembly-*.adoc",
-    "build:generated-modules": "asciidoctor --failure-level=WARN --trace --template-dir=templates --destination-dir=tmp/generated/modules tmp/post-splitter/modules/*/*.adoc",
+    "build:generated-modules": "asciidoctor --failure-level=WARN --trace --template-dir=templates --destination-dir=tmp/generated/modules tmp/post-splitter/modules/*/*.adoc","generate:attributes": "node ./utils/attributes.js",
```
LightGuard, line 64: Maybe I'm missing something, but it looks like you're a step per line for readablity. This one should probably follow suit.

LightGuard (commented):
Everything looks good to me and the checks passed.

.build/package.json
pmuir: Thanks - yes, I just realised that. I also realised I can get rid of the `pom.xml` and just use the mvn command line to get the dependency directly.

.build/package.json
```diff
@@ -56,15 +57,17 @@
     "start": "webpack-cli serve",
     "start:dev": "webpack serve --hot --color --progress --config webpack.dev.js",
     "clean": "rimraf dist tmp && gh-pages-clean",
-    "generate:attributes": "node ./utils/attributes.js",
-    "generate:modular-docs-sources": "node ./utils/modular-docs.js",
     "commitandpush:modular-docs": "gh-pages -d ./tmp/post-splitter -b modular-docs",
     "publish:modular-docs": "npm run generate:modular-docs-sources && npm run commitandpush:modular-docs",
     "fetch:templates": "node ./utils/templates.js",
+    "fetch:splitter": "mvn org.apache.maven.plugins:maven-dependency-plugin:3.2.0:copy -Dartifact=io.github.lightguard.documentation:asciidoc-splitter:1.5.3:jar:jar-with-dependencies -DoutputDirectory=tmp/binaries/splitter",
```
LightGuard, line 63: Never used Maven this way, that's handy!

LightGuard (approved):

PR #371, opened 2021-11-30
Fix metrics content for now.

sterobin:
Fixes bug reported in [MGDSTRM-6388](https://issues.redhat.com/browse/MGDSTRM-6388).


bhardesty (approved):
Looks good!

PR #372, opened 2021-12-02
Add missing dependency

craicoverflow:
This dependency was missing so I could not run the guides locally:

```shell
$ webpack serve --hot --color --progress --config webpack.dev.js
[webpack-cli] Unable to load '@webpack-cli/serve' command
[webpack-cli] TypeError: options.forEach is not a function
```

### DIFF
--- a/.build/package-lock.json
+++ b/.build/package-lock.json
@@ -19,6 +19,7 @@
         "yaml": "1.10.2"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


wtrocki (approved):

craicoverflow:
> We need to regenerate shrinkwrap file. Build will not allow us to do update without it

I was using yarn, so it did not update.

PR #375, opened 2021-12-09
New metrics doc

sterobin:
Jameel and Tom, ready for your SME review. Doc preview linked below. **Please note especially the tags `@SME` for questions to you.**

Links:
* [Doc preview](http://file.rdu.redhat.com/~sterobin/MGDSTRM-6077/README.html)
* [Jira](https://issues.redhat.com/browse/MGDSTRM-6077)
* [Corresponding operator PR](https://github.com/redhat-developer/app-services-operator/pull/295)


metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+This procedure follows the https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/additional-scrape-config.md#additional-scrape-configuration[Additional Scrape Configuration] method defined by Prometheus for integrating third-party metrics.
+
+.Prerequisites
+* You have access to a Kafka instance that contains topics in {product}. For more information about access management in {product}, see {base-url}{access-mgmt-url}[_Managing account access in {product-long}_^].
+* You have the generated credentials for your service account that has access to the Kafka instance. To regenerate and re-copy the credentials, use the *Service Accounts* page in the {product} web console to find your service account and update the credentials. For more information about service accounts, see {base-url}{getting-started-url}#proc-creating-service-account_getting-started[_Getting started with {product-long}_^].
+* You or other users have produced and consumed messages in the Kafka topics. For example, to use Kafka scripts to produce and consume messages, see {base-url}{kafka-bin-scripts-url}[_Configuring and connecting Kafka scripts with {product-long}_^].
```
JameelB, line 145: I don't really see this as a prerequisite of configuring the scrape endpoint in Prometheus. This can still be set up even if the Kafka instance hasn't been used yet (no connections, no topics, etc.). The metrics around those may not appear until then but the scrape endpoint will still be up and running. Once Kafka is used, those metrics should then appear/result updated.

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+  - targets: ["api.openshift.com"]
+  scheme: "https"
+  metrics_path: "/api/kafkas_mgmt/v1/kafkas/__<kafka_instance_id>__/metrics/federate"
+  oauth2:
+    client_id: "__<client_id>__"
+    client_secret: "__<client_secret>__"
+    token_url: "https://identity.api.openshift.com/auth/realms/rhoas/protocol/openid-connect/token"
+----
+--
+. Create a Kubernetes secret that contains the new `kafka-federate.yaml` file, as shown in the following example. Replace `<namespace>` with the value of your own Kubernetes namespace.
```
JameelB, line 185: Might be good to note that this should be the same namespace where their Prometheus instance is deployed.

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+* You have the generated credentials for your service account that has access to the Kafka instance. To regenerate and re-copy the credentials, use the *Service Accounts* page in the {product} web console to find your service account and update the credentials. For more information about service accounts, see {base-url}{getting-started-url}#proc-creating-service-account_getting-started[_Getting started with {product-long}_^].
+* You or other users have produced and consumed messages in the Kafka topics. For example, to use Kafka scripts to produce and consume messages, see {base-url}{kafka-bin-scripts-url}[_Configuring and connecting Kafka scripts with {product-long}_^].
+* You've installed a Prometheus instance. For installation instructions, see https://prometheus.io/docs/prometheus/latest/getting_started/[Getting Started] in the Prometheus documentation.
+* You've installed the https://kubernetes.io/docs/tasks/tools/#kubectl[kubectl] Kubernetes command-line tool.
+
+.Procedure
+. In your Prometheus instance, create the following `kafka-federate.yaml` file. Replace the values with your own Kafka instance and credential information.
```
JameelB, line 150:
```suggestion
. Create the following `kafka-federate.yaml` file. Replace the values with your own Kafka instance and credential information.
```

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+  - targets: ["api.openshift.com"]
+  scheme: "https"
+  metrics_path: "/api/kafkas_mgmt/v1/kafkas/__<kafka_instance_id>__/metrics/federate"
+  oauth2:
+    client_id: "__<client_id>__"
+    client_secret: "__<client_secret>__"
+    token_url: "https://identity.api.openshift.com/auth/realms/rhoas/protocol/openid-connect/token"
+----
+
+Alternatively, you can add this information to your Prometheus configuration file, as shown in the following example. For more information about this method, see https://prometheus.io/docs/prometheus/latest/configuration/configuration/#configuration-file[Configuration File] in the Prometheus documentation.
```
JameelB, line 169: Might be good to separate this out from these steps? Once you do this, you don't really need to do Step 2 and 3 below.

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+== Supported metrics in {product}
+
+{product} supports the following metrics for Kafka instances and topics. In the {product} web console, a subset of these metrics is displayed in the *Dashboard* page of each Kafka instance.
+
+Broker metrics::
++
+--
+* `kafka_broker_quota_softlimitbytes`: Maximum amount of broker storage before producers are throttled
+
+* `kafka_broker_quota_totalstorageusedbytes`: Total storage used by the broker
```
tombentley, line 67: It's not a docs comment, but it's the first time I've really see how we're exposing metrics to users of the service... The naming of these could be problematic; they are not Kafka metrics. AFAIK they come from the [quota plugin](https://github.com/strimzi/kafka-quotas-plugin/blob/main/src/main/java/io/strimzi/kafka/quotas/StaticQuotaCallback.java#L160), where the name makes clear that they're a Strimzi, rather than Kafka thing. IMHO it is at least confusing to give these the `kafka_` prefix when they don't come from Kafka. It doesn't have to use a `strimzi` prefix, but perhaps we could use some other prefix. 

Slightly more generally, if there's a pattern to how we're using `:` in names it's not clear to me: Why is it `kafka_topic:kafka_topic_partitions...` but `kafka_broker_quota_...`?

@pmuir have we really thought about these names?

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+As a developer or administrator, you can view metrics in {product} to visualize the performance and data usage for Kafka instances and topics that you have access to. You can view metrics directly in the {product} web console, or use the metrics API endpoint provided by {product} to import the data into your own metrics monitoring tool, such as Prometheus.
+
+//Additional line break to resolve mod docs generation error, not sure why. Leaving for now. (Stetson, 20 May 2021)
+
+[id="ref-supported-metrics_{context}"]
+== Supported metrics in {product}
+
+{product} supports the following metrics for Kafka instances and topics. In the {product} web console, a subset of these metrics is displayed in the *Dashboard* page of each Kafka instance.
+
+Broker metrics::
```
tombentley, line 62: Do we need to describe the labels each of these metrics will have? I think it would make it a little clearer for some of the metrics, like `kafka_topic:kafka_topic_partitions:count` (where I assume there's a series for each topic).

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+* `kafka_topic:kafka_topic_partitions:count`: Number of partitions in a topic *<@SME: This correct?>*
+
+* `kafka_topic:kafka_topic_partitions:sum`: *<@SME: Description? I don't see the diff between this and "partitions:count".>*
+--
+
+Scrape metrics::
++
+--
+*<SME: Description for these? I can kind of guess on a couple, but not clear.>*
```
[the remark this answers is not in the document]
sterobin: Ah, then no. Removing. Thanks.

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+--
+* `kafka_broker_quota_softlimitbytes`: Maximum amount of broker storage before producers are throttled
+
+* `kafka_broker_quota_totalstorageusedbytes`: Total storage used by the broker
+
+* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions across all topics in the cluster
+
+* `kafka_controller_kafkacontroller_offline_partitions_count`: Number of partitions without an active leader and that are therefore not writable or readable
+
+* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: *<@SME: Description?>*
```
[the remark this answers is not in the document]
sterobin: Ack. But for this and the outgoing rate, I've said "Incoming byte rate for **the broker** in the last five minutes" because all of these metrics seem to be specifically for "server" and "broker" settings, except the topic ones. Hence all other similar metrics refer to broker and are under a "Broker" heading. Also Confluent classifies all of these as metrics for the broker. So is it correct to say that for consistency? Not sure when I should say "Kafka instance" versus stick to saying "broker" for consistency and accuracy.

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+* `kafka_namespace:haproxy_server_bytes_out_total:rate5m`: *<@SME: Description?>*
```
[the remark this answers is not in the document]
sterobin: Ditto

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+Topic metrics::
++
+--
+* `kafka_server_brokertopicmetrics_bytes_in_total`: Number of incoming bytes to topics in the instance
+
+* `kafka_server_brokertopicmetrics_bytes_out_total`: Number of outgoing bytes from topics in the instance
+
+* `kafka_server_brokertopicmetrics_messages_in_total`: Number of messages per second received by one or more topics in the instance
+
+* `kafka_topic:kafka_log_log_size:sum`: Log size for each partition in a topic *<@SME: This correct?>*
```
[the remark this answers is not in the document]
sterobin: Is it "Log size for each topic" meaning it's a per-topic metric, or is it "Total log size of all topics in the instance" as a sum? I'm confused by the `sum` in the metric name.

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+
+* `kafka_topic:kafka_topic_partitions:count`: Number of partitions in a topic *<@SME: This correct?>*
+
+* `kafka_topic:kafka_topic_partitions:sum`: *<@SME: Description? I don't see the diff between this and "partitions:count".>*
```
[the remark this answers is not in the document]
sterobin: Interesting. Then what's the difference now between these two metrics, which sound the same to me? One is a broker metric, the other a topic one. Is the first one (broker) for all partitions across all instances, aka, across the cluster, and the second one (topic) just the partitions for all topics in one instance in that cluster? Not super clear because I've been hearing the Kafka world refer to a Kafka instance and Kafka cluster as the same thing, synonymous.

```
`kafka_controller_kafkacontroller_global_partition_count`: Number of partitions across all topics in the cluster

`kafka_topic:kafka_topic_partitions:sum`: Number of partitions across all topics in the instance
```

metrics-monitoring/README.adoc
sterobin: Yes, great point. For both this and the other procedure, I've removed that prereq and moved it down into the actual steps as a mention of expected behavior when users first create new instances and topics. Thanks.

metrics-monitoring/README.adoc
sterobin: Agreed, done.

metrics-monitoring/README.adoc
sterobin: Yeah, I went back and forth on this. Separating it out felt like too much emphasis on the alternative route, and mentioning it at the end of the proc started to pile up the end of the proc with the info that's already there. But I definitely see the lack of clarity. So I've turned it into a note instead and clarified that doing that is an alternative that then negates the remaining steps. It also keeps the same setup file content together but for the two options.

Anyway, have a look at my new version of that and see what you think.

metrics-monitoring/README.adoc
sterobin: @tombentley Ack. Yes, agreed that that's a convo outside the scope of this PR and one worth having. We'll move forward with whatever naming is currently a reality. Once they are updated, anyone is welcome to update the metrics names accordingly in the docs at that point. Thanks.

metrics-monitoring/README.adoc
sterobin: That would be nice. Reason we've just listed them is that this is what the user will have to choose from when they set them up initially. Then when they actually appear, they new label is shown, which I agree might not always be super intuitive like you say. If those labels are pretty certain, I agree would be nice to list them, like in parentheses or something for each. But I'd need some help with that yet again. Just getting this list of metrics required help from Pete.

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
JameelB: it's not a placeholder but rather how the metric is grouped/aggregated. The metrics with `:` are recording rules and the naming convention was suggested by @david-martin

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,225 @@
...
+--
+* `kafka_server_brokertopicmetrics_bytes_in_total`: Number of incoming bytes to topics in the instance
+
+* `kafka_server_brokertopicmetrics_bytes_out_total`: Number of outgoing bytes from topics in the instance
+
+* `kafka_server_brokertopicmetrics_messages_in_total`: Number of messages per second received by one or more topics in the instance
+
+* `kafka_topic:kafka_log_log_size:sum`: Log size for each partition in a topic *<@SME: This correct?>*
+
+* `kafka_topic:kafka_topic_partitions:count`: Number of partitions in a topic *<@SME: This correct?>*
```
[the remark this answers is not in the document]
sterobin: Ack, thanks.

metrics-monitoring/README.adoc
sterobin: But they can't just create that file anywhere, has to be in the prometheus install folder or in the location of the installed instance, right? That's what I've found when I've gone through this. If that's not the case, we'd still need to tell them where to put that new file, otherwise they might think, "Create it where exactly?"

metrics-monitoring/README.adoc
tombentley: OK, thanks. It's worth thinking about because this is API and people will build their own reporting and alerting around these things.

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
tombentley: Now it makes some sense, thanks! But the fact that they're implemented by recording rules is an impl detail as far as the user is concerned. I think it would be better to have a naming schema that hid that detail.

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
tombentley: The sum will be because the recording rule is doing an aggregation. Again, that's an impl detail.

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
tombentley: I think the difference will be in the labels (which is what prompted me to ask where adding them would make things clearer).

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
JameelB: It's your first statement there: `"Log size for each topic" meaning it's a per-topic metric`. The `sum` is the operator used when aggregating this metric but it's grouped per `kafka_topic`.

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
JameelB: I'm not too familiar with the `kafka_controller_kafkacontroller_global_partition_count` metric but from looking at its results, it's grouped per broker whereas the `kafka_controller_kafkacontroller_global_partition_count` is partitions across the instance.

metrics-monitoring/README.adoc
JameelB: That file is only really used to generate the secret in step 2 so you should be able to create it anywhere in your local directory as long as you update the path in the kubectl cmd:

```
kubectl create secret generic additional-scrape-configs --from-file=<path-to-kafka-federate.yaml> --dry-run -o yaml \
kubectl apply -f - -n <namespace>
```

metrics-monitoring/README.adoc
pmuir: @tombentley can you raise on the mas-devel list? I'm also not overly happy with the names of the metrics (they are hard to follow) but you probably have more insight into what they should be than I do.

metrics-monitoring/README.adoc
JameelB: makes sense, what you have there now looks good to me :+1:

metrics-monitoring/README.adoc
JameelB: Labels should be constant, they're defined [here](https://github.com/bf2fc6cc711aee1a0c2a/kas-fleet-manager/blob/main/internal/kafka/constants/metrics.go). The `VariableLabels` contains the labels returned for each metric.

metrics-monitoring/README.adoc
pmuir: This only applies if you are using the prometheus operator as well.

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
sterobin: @JameelB Ack. I see. I've just updated this and several other metric descriptions to say "in the instance" etc., instead of "in the broker", because there could be multiple brokers or other components in the instance involved for a metric. I think that broadens the scope a bit here and elsewhere to avoid the innacuracy.

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
sterobin: Got it, thanks.

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
sterobin: I see. I've changed as follows. This work?

```
`kafka_controller_kafkacontroller_global_partition_count`: Number of partitions per broker in the instance

`kafka_topic:kafka_topic_partitions:sum`: Number of partitions across all topics in the instance
```

metrics-monitoring/README.adoc
sterobin: @JameelB ack, thanks.

metrics-monitoring/README.adoc
sterobin: @JameelB @pmuir Pete's last comment about it applying only to the operator got me thinking and I spoke with Pete and learned that really the configuration file method is the more preferred and simpler method, and the additional config method is the alternative if needed for the operator scenarios. So I've actually now restructured this prometheus procedure to essentially reverse the preference.

Can you two have another look at the new procedure and see what you think now? **Use the same doc preview link in the PR description above and refresh the page.**

metrics-monitoring/README.adoc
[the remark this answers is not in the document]
JameelB: yep that makes sense :+1:

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,210 @@
...
+Topic metrics::
++
+--
+* `kafka_server_brokertopicmetrics_bytes_in_total`: Number of incoming bytes to topics in the instance
+
+* `kafka_server_brokertopicmetrics_bytes_out_total`: Number of outgoing bytes from topics in the instance
+
+* `kafka_server_brokertopicmetrics_messages_in_total`: Number of messages per second received by one or more topics in the instance
+
+* `kafka_topic:kafka_log_log_size:sum`: Log size of all topics in the instance
```
JameelB, line 95:
```suggestion
* `kafka_topic:kafka_log_log_size:sum`: Log size for each topic in the instance
```

metrics-monitoring/README.adoc
JameelB: Latest update makes sense and looks good to me :)

metrics-monitoring/README.adoc
sterobin: Done, thanks.

metrics-monitoring/README.adoc
sterobin: Thanks @JameelB !

metrics-monitoring/README.adoc
bhardesty: I also found the list of metrics difficult to grok, especially when they all appear in a long list. @sterobin you might consider switching the order and listing the definition first (since that's the part that most readers care about the most), and then the technical name of the metric.

So, for example instead of this:

```
* `kafka_broker_quota_softlimitbytes`: Maximum amount of broker storage before producers are throttled
```
Do something like this:

```
* Maximum amount of broker storage before producers are throttled (`kafka_broker_quota_softlimitbytes`) 
```

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,210 @@
...
+  metrics_path: "/api/kafkas_mgmt/v1/kafkas/__<kafka_instance_id>__/metrics/federate"
+  oauth2:
+    client_id: "__<client_id>__"
+    client_secret: "__<client_secret>__"
+    token_url: "https://identity.api.openshift.com/auth/realms/rhoas/protocol/openid-connect/token"
+----
+
+The new scrape target becomes available after the configuration has reloaded.
+
+You can view your collected metrics in the Prometheus expression browser at `http://__<host>__:__<port>__/graph`, or integrate your Prometheus data source with a data-graphing tool such as Grafana. For information about Prometheus metrics in Grafana, see https://prometheus.io/docs/visualization/grafana/[Grafana Support for Prometheus] in the Grafana documentation.
```
bhardesty, line 152: Should this be an optional step 2?

metrics-monitoring/README.adoc
```diff
@@ -0,0 +1,210 @@
...
+----
+
+.Example command to create and apply a Kubernetes secret
+[source,subs="+quotes"]
+----
+kubectl create secret generic additional-scrape-configs --from-file=__<~/kafka-federate.yaml>__ --dry-run -o yaml \
+kubectl apply -f - -n __<namespace>__
+----
+
+.Example Prometheus custom resource with new secret
```
bhardesty, line 183: Are we still supposed to capitalize "Custom Resource"?

bhardesty (approved):
This looks really good; I just made several small, optional comments/suggestions.

metrics-monitoring/README.adoc
sterobin: Right, so the reason I listed the metrics like this is that this is how users will be keying them when they look them up. Meaning, they go to Prometheus or wherever and then open the list of available metrics that they can choose from to set up their metrics dashboard. So these are the metrics they'll need to choose from and will need to know what each means. So they're basically metrics entries that correlate with what they see in Prometheus. If I put definitions first, it will bury the things that they're comparing against for the information.

Regarding the confusingness of the metric names, that's certainly an issue, but a separate one. The names are definitely system-centered and not user-centered. That was Tom's

metrics-monitoring/README.adoc
sterobin: That's a good question. I guess imo no, not an optional step, just saying where they can see the results after they've done the configuration, which is what the proc is really about. So kind of an end result I guess. I'll think on that.

metrics-monitoring/README.adoc
sterobin: Not that I recall. We agreed just to follow the OpenShift approach and capitalize only Operator: https://github.com/openshift/openshift-docs/blob/main/contributing_to_docs/doc_guidelines.adoc#api-object-formatting

metrics-monitoring/README.adoc
david-martin: `Slightly more generally, if there's a pattern to how we're using : in names it's not clear to me: Why is it kafka_topic:kafka_topic_partitions... but kafka_broker_quota_...?`

@tombentley I would expect any recording rules to follow the convention in https://prometheus.io/docs/practices/rules/#naming-and-aggregation

b1zzu (approved):
I manually connected a Kafka instance to a Prometheus instance following this doc successfully and validated the rest. LGTM

PR #383, opened 2022-01-05
[WIP] Add new tutorial to service binding guide for binding Quarkus to OpenShift Service Registry

jbyrne-redhat:
**Notes for reviewers**
 
 - Add new conceptual content describing file-based binding for OpenShift Service Registry. [HTML preview](http://file.emea.redhat.com/~jbyrne/j6267/README.html#con-about-service-binding_service-binding).
-  Update `rhoas cluster connect` procedure for Service Registry (in addition to Kafka). Add steps for RBAC. [HTML preview](http://file.emea.redhat.com/~jbyrne/j6267/README.html#proc-connecting-kafka-registry-instance-to-openshift-cluster_service-binding).
-  Add tutorial for binding new Quarkus application to Service Registry (in addition to Kafka). [HTML preview](http://file.emea.redhat.com/~jbyrne/j6267/README.html#con-kafka-registry-binding-quarkus-application-using-cli_service-binding).

PR #389, opened 2022-01-26
[WIP] [Reopened] Add new tutorial to service binding guide for binding Quarkus to OpenShift Service Registry using the CLI

jbyrne-redhat:
#383 was prematurely merged. This PR reopens it for fresh review.

PR #390, opened 2022-01-28
"[WIP] Add new tutorial to service binding guide for binding Quarkus to OpenShift Service Registry"

wtrocki:
Reverts redhat-developer/app-services-guides#383


jbyrne-redhat (approved):
Thanks, @wtrocki. Makes sense to revert for now. I have #389 to make the originally-intended changes when the time is right.

PR #397, opened 2022-02-09
Update metrics descriptions

sterobin:
Expanded metrics descriptions per feedback from Mickael Maison in [google doc](https://docs.google.com/document/d/1V_E9G-SULu_zZCNlQ2FF-3TgunOZLatLNMiKsnnAIwA/edit).

[Doc preview](http://file.rdu.redhat.com/~sterobin/MGDSTRM-metrics-monitoring/README.html#ref-supported-metrics_monitoring-metrics)
[Jira](https://issues.redhat.com/browse/MGDSTRM-7243)

### DIFF
--- a/docs/kafka/metrics-monitoring-kafka/README.adoc
+++ b/docs/kafka/metrics-monitoring-kafka/README.adoc
@@ -66,44 +66,48 @@ As a developer or administrator, you can view metrics in {product-kafka} to visu
 [role="_abstract"]
 {product-kafka} supports the following metrics for Kafka instances and topics. In the {product-kafka} web console, a subset of these metrics is displayed in the *Dashboard* page of each Kafka instance.
 
+Cluster metrics::
++
+--
+* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: Number of incoming bytes for the cluster in the last five minutes. This metric represents all the data that producers are sending to topics in the cluster. The maximum incoming byte rate is 30 MB per second.
+
+* `kafka_namespace:haproxy_server_bytes_out_total:rate5m`: Number of outgoing bytes for the cluster in the last five minutes. This metric represents all the data that producers are receiving from topics in the cluster. The maximum outgoing byte rate is 30 MB per second.
+
+* `kafka_namespace:kafka_server_socket_server_metrics_connection_count:sum`: Number of current client connections to the cluster. Kafka clients use persistent connections to interact with brokers in the cluster. For example, a consumer holds a connection to each broker it is receiving data from and a connection to its group coordinator. The maximum number of active connections is 2000.
+
+* `kafka_namespace:kafka_server_socket_server_metrics_connection_creation_rate:sum`: Number of client connection creations per second for the cluster. Kafka clients use persistent connections to interact with brokers in the cluster. A constant high number of connection creations might indicate a client issue. The maximum connection creation rate is 100 per second.
+
+* `kafka_topic:kafka_topic_partitions:count`: Number of topics in the cluster. This does not include internal Kafka topics, such as `__consumer_offsets` and `__transaction_state`. The maximum number of topics is 1000.
+
+* `kafka_topic:kafka_topic_partitions:sum`: Number of partitions across all topics in the cluster. This does not include internal Kafka topics, such as `__consumer_offsets` and `__transaction_state`. The maximum number of partitions is 1000.
+--
+
 Broker metrics::
 +
 --
-* `kafka_broker_quota_softlimitbytes`: Maximum amount of broker storage before producers are throttled
+* `kafka_broker_quota_softlimitbytes`: Maximum amount of storage, in bytes, for this broker before producers are throttled. When this limit is reached, the broker starts throttling producers to prevent them from sending additional data. The maximum storage in the broker is 1000 GiB.
 
-* `kafka_broker_quota_totalstorageusedbytes`: Total storage used by the broker
+* `kafka_broker_quota_totalstorageusedbytes`: Amount of storage, in bytes, that is currently used by partitions in the broker. The storage usage depends on the number and retention configurations of the partitions. This metric must stay below the `kafka_broker_quota_softlimitbytes` metric setting.
 
-* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions for each broker in the instance
+* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions in the cluster. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`. This count includes partitions from internal Kafka topics, such as `__consumer_offsets` and `__transaction_state`. The maximum number of partitions is 1000.
 
-* `kafka_controller_kafkacontroller_offline_partitions_count`: Number of partitions without an active leader and that are therefore not writable or readable
+* `kafka_controller_kafkacontroller_offline_partitions_count`: Number of partitions in the cluster that are currently offline. Offline partitions cannot be used by clients for producing or consuming data. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`.
 
-* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: Incoming byte rate for the instance in the last five minutes
+* `kubelet_volume_stats_available_bytes`: Amount of disk space, in bytes, that is available in the broker.
 
-* `kafka_namespace:haproxy_server_bytes_out_total:rate5m`: Outgoing byte rate for the instance in the last five minutes
-
-* `kafka_namespace:kafka_server_socket_server_metrics_connection_count:sum`: Number of client connections to the instance
-
-* `kafka_namespace:kafka_server_socket_server_metrics_connection_creation_rate:sum`: Number of attempted client connections per second for the instance
-
-* `kubelet_volume_stats_available_bytes`: Available disk space in the instance
-
-* `kubelet_volume_stats_used_bytes`: Used disk space in the instance
+* `kubelet_volume_stats_used_bytes`: Amount of disk space, in bytes, that is currently used in the broker.
 --
 
 Topic metrics::
 +
 --
-* `kafka_server_brokertopicmetrics_bytes_in_total`: Number of incoming bytes to topics in the instance
+* `kafka_server_brokertopicmetrics_bytes_in_total`: Number of incoming bytes to topics in the instance.
 
-* `kafka_server_brokertopicmetrics_bytes_out_total`: Number of outgoing bytes from topics in the instance
+* `kafka_server_brokertopicmetrics_bytes_out_total`: Number of outgoing bytes from topics in the instance.
 
-* `kafka_server_brokertopicmetrics_messages_in_total`: Number of messages per second received by one or more topics in the instance
+* `kafka_server_brokertopicmetrics_messages_in_total`: Number of messages per second received by one or more topics in the instance.
 
-* `kafka_topic:kafka_log_log_size:sum`: Log size of each topic in the instance
-
-* `kafka_topic:kafka_topic_partitions:count`: Number of topics in the instance
-
-* `kafka_topic:kafka_topic_partitions:sum`: Number of partitions across all topics in the instance
+* `kafka_topic:kafka_log_log_size:sum`: Log size of each topic and replica, in bytes, across all brokers in the cluster.
 --
 
 [id="proc-viewing-metrics_{context}"]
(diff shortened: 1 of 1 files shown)


metrics-monitoring/README.adoc
```diff
@@ -63,44 +63,48 @@ As a developer or administrator, you can view metrics in {product} to visualize
...
 
-* `kubelet_volume_stats_used_bytes`: Used disk space in the instance
+* `kafka_controller_kafkacontroller_offline_partitions_count`: Number of partitions in the cluster that are currently offline. Offline partitions cannot be used by clients for producing or consuming data. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`.
+
+* `kubelet_volume_stats_available_bytes`: Amount of disk space, in bytes, that is available in the broker.
+
+* `kubelet_volume_stats_used_bytes`: Amount of disk space, in bytes, that is currently used in the broker.
 --
 
 Topic metrics::
```
sterobin, line 98: @Mickael, I realize that you actually split the first 3 of these off as a separate "topic broker metric" section, but I'm not sure if users will know what "topic broker" metrics are. I think people more likely know what broker metrics, topic metrics, and cluster metrics are, respectively. Thought it might create more confusion splitting it that way, but let me know what you think.

metrics-monitoring/README.adoc
```diff
@@ -63,44 +63,48 @@ As a developer or administrator, you can view metrics in {product} to visualize
...
+* `kafka_broker_quota_totalstorageusedbytes`: Amount of storage, in bytes, that is currently used by partitions in the broker. The storage usage depends on the number and retention configurations of the partitions. This metric must stay below the `kafka_broker_quota_softlimitbytes` metric setting.
 
-* `kubelet_volume_stats_available_bytes`: Available disk space in the instance
+* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions in the cluster. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`. This count includes partitions from internal Kafka topics, such as `__consumer_offsets` and `__transaction_state`. The maximum number of partitions is 1000.
 
-* `kubelet_volume_stats_used_bytes`: Used disk space in the instance
+* `kafka_controller_kafkacontroller_offline_partitions_count`: Number of partitions in the cluster that are currently offline. Offline partitions cannot be used by clients for producing or consuming data. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`.
+
```
sterobin, line 92: Agreed that the following two metrics are basically repeats of other metrics. Not sure what we should do here in docs if they're that way in the display.

metrics-monitoring/README.adoc
```diff
@@ -63,44 +63,48 @@ As a developer or administrator, you can view metrics in {product} to visualize
...
 
-* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions for each broker in the instance
+* `kafka_namespace:kafka_server_socket_server_metrics_connection_count:sum`: Number of current client connections to the cluster. Kafka clients use persistent connections to interact with brokers in the cluster. For example, a consumer holds a connection to each broker it is receiving data from and a connection to its group coordinator. The maximum number of active connections is 2000.
 
-* `kafka_controller_kafkacontroller_offline_partitions_count`: Number of partitions without an active leader and that are therefore not writable or readable
+* `kafka_namespace:kafka_server_socket_server_metrics_connection_creation_rate:sum`: Number of client connection creations per second for the cluster. Kafka clients use persistent connections to interact with brokers in the cluster. A constant high number of connection creations might indicate a client issue. The maximum connection creation rate is 100 per second.
 
-* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: Incoming byte rate for the instance in the last five minutes
```
sterobin, line 77: @Mickael, have another look at the following two metrics. They were very confusing to me last time we wrote them because they basically seemed the same. Then your new definition of each was literally identical. So I've kept some of the old definition and some of your new to differentiate. Let me know what you think.

metrics-monitoring/README.adoc
mimaison: Yes they are basically repeats. They are computed from different components though so if you plot them they are close but not exactly the same. Maybe we can use the same descriptions?

metrics-monitoring/README.adoc
mimaison: These metrics are per broker and per topic, they have both labels. I thought mentioning the labels associated with each metric was useful so I grouped them this way. If you find it confusing, you can keep them in the "topic" category.

metrics-monitoring/README.adoc
mimaison: yes that's fine

metrics-monitoring/README.adoc
sterobin: Hm, interesting. Well I'm not sure what to do here because which one do we tell users to use for which? If I point out for clarity that they are essentially the same, then users will want to know when to use which or if it genuinely doesn't make a difference. Do you know if it's true that users can use either metric as equally valid for the same purpose? If so, I suppose I could just mention that in the description. That way at least they're not wondering if it's their imagination.

sterobin: Hm, okay. So should I just say "This metric is similar to the X metric" ?

metrics-monitoring/README.adoc
```diff
@@ -63,44 +63,48 @@ As a developer or administrator, you can view metrics in {product} to visualize
...
 
-* `kafka_namespace:kafka_server_socket_server_metrics_connection_creation_rate:sum`: Number of attempted client connections per second for the instance
+* `kafka_broker_quota_totalstorageusedbytes`: Amount of storage, in bytes, that is currently used by partitions in the broker. The storage usage depends on the number and retention configurations of the partitions. This metric must stay below the `kafka_broker_quota_softlimitbytes` metric setting.
 
-* `kubelet_volume_stats_available_bytes`: Available disk space in the instance
+* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions in the cluster. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`. This count includes partitions from internal Kafka topics, such as `__consumer_offsets` and `__transaction_state`. The maximum number of partitions is 1000. This metric is similar to the `kafka_topic:kafka_topic_partitions:sum` cluster metric.
```
sterobin, line 89: @mimaison I've added the last line, "This metric is similar to ..." That work?

metrics-monitoring/README.adoc
```diff
@@ -63,44 +63,48 @@ As a developer or administrator, you can view metrics in {product} to visualize
...
 
-* `kubelet_volume_stats_used_bytes`: Used disk space in the instance
+* `kafka_controller_kafkacontroller_offline_partitions_count`: Number of partitions in the cluster that are currently offline. Offline partitions cannot be used by clients for producing or consuming data. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`.
+
+* `kubelet_volume_stats_available_bytes`: Amount of disk space, in bytes, that is available in the broker.
+
+* `kubelet_volume_stats_used_bytes`: Amount of disk space, in bytes, that is currently used in the broker. This metric is similar to the `kafka_broker_quota_totalstorageusedbytes` broker metric.
```
sterobin, line 95: @mimaison Same here, I've added the last line, "This metric is similar to ..." That work? Any others I should call out as similar?

metrics-monitoring/README.adoc
```diff
@@ -63,44 +63,48 @@ As a developer or administrator, you can view metrics in {product} to visualize
...
 
-* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions for each broker in the instance
+* `kafka_namespace:kafka_server_socket_server_metrics_connection_count:sum`: Number of current client connections to the cluster. Kafka clients use persistent connections to interact with brokers in the cluster. For example, a consumer holds a connection to each broker it is receiving data from and a connection to its group coordinator. The maximum number of active connections is 2000.
 
-* `kafka_controller_kafkacontroller_offline_partitions_count`: Number of partitions without an active leader and that are therefore not writable or readable
+* `kafka_namespace:kafka_server_socket_server_metrics_connection_creation_rate:sum`: Number of client connection creations per second for the cluster. Kafka clients use persistent connections to interact with brokers in the cluster. A constant high number of connection creations might indicate a client issue. The maximum connection creation rate is 100 per second.
```
mimaison, line 75: What about changing `indicate a client issue` to `indicate a client-side issue`?

metrics-monitoring/README.adoc
```diff
@@ -63,44 +63,48 @@ As a developer or administrator, you can view metrics in {product} to visualize
...
 
-* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: Incoming byte rate for the instance in the last five minutes
+* `kafka_topic:kafka_topic_partitions:count`: Number of topics in the cluster. This does not include internal Kafka topics, such as `__consumer_offsets` and `__transaction_state`. The maximum number of topics is 1000.
 
-* `kafka_namespace:haproxy_server_bytes_out_total:rate5m`: Outgoing byte rate for the instance in the last five minutes
+* `kafka_topic:kafka_topic_partitions:sum`: Number of partitions across all topics in the cluster. This does not include internal Kafka topics, such as `__consumer_offsets` and `__transaction_state`. The maximum number of partitions is 1000.
```
mimaison, line 79: What about changing `include internal Kafka topics` to `include partitions from internal Kafka topics`?

metrics-monitoring/README.adoc
```diff
@@ -63,44 +63,48 @@ As a developer or administrator, you can view metrics in {product} to visualize
 [role="_abstract"]
 {product} supports the following metrics for Kafka instances and topics. In the {product} web console, a subset of these metrics is displayed in the *Dashboard* page of each Kafka instance.
 
-Broker metrics::
+Cluster metrics::
 +
 --
-* `kafka_broker_quota_softlimitbytes`: Maximum amount of broker storage before producers are throttled
+* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: Number of incoming bytes for the cluster in the last five minutes. This metric represents all the data that producers are sending to topics in the cluster. The maximum incoming byte rate is 30 MB per second.
```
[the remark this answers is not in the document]
sterobin: Good catch, yes, agreed. Done.

metrics-monitoring/README.adoc
sterobin: Just trimming. Saying "client-side issue" is a bit colloquial for official tech docs, when it's meaning an issue with the client, so aka, "client issue" communicates the same thing to me but less casual.

metrics-monitoring/README.adoc
```diff
@@ -63,44 +63,48 @@ As a developer or administrator, you can view metrics in {product} to visualize
...
 
-* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: Incoming byte rate for the instance in the last five minutes
+* `kafka_topic:kafka_topic_partitions:count`: Number of topics in the cluster. This does not include internal Kafka topics, such as `__consumer_offsets` and `__transaction_state`. The maximum number of topics is 1000.
 
-* `kafka_namespace:haproxy_server_bytes_out_total:rate5m`: Outgoing byte rate for the instance in the last five minutes
+* `kafka_topic:kafka_topic_partitions:sum`: Number of partitions across all topics in the cluster. This does not include internal Kafka topics, such as `__consumer_offsets` and `__transaction_state`. The maximum number of partitions is 1000.
+--
 
-* `kafka_namespace:kafka_server_socket_server_metrics_connection_count:sum`: Number of client connections to the instance
+Broker metrics::
```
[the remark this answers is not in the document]
sterobin: Right, I didn't included them because I wondered if it was really necessary. I think it's pretty clear that anything under "brokers" is per broker, under "topics" per topic, etc. So the line just didn't provide any really new info in my mind. However, the thing it does provided is the label bit, but I'm not sure again if that's necessary given that the metrics are there listed, and also there was the issue you pointed out in the "cluster" section that the label is missing. So I don't want to point out to users what the label is for each section and then not point that out for cluster, since it's missing. Seemed odd and would likely lead to questions and look like we just left that out by accident. 

So in light of all that, I thought it would avoid the issues by just nixing those lines. Happy to chat more if you disagree or if I've misunderstood.

metrics-monitoring/README.adoc
mimaison: That makes sense, thanks

mimaison (approved):
LGTM

PR #398, opened 2022-02-10
fix: proposed file structure per product

wtrocki:
FYI this is early version of PR that targets to setup different repository structure. I have put some proposal as I need some starting structure for automation to work on. 

## Structure

- docs folder contains all documentation - separates codebase
- kafka,registry etc. folders in the docs represents service specific documentation 
- There is no common folder - we would implicitly include other relevant guides if needed.

@sterobin WDYT

FYI
@redhat-developer/app-services-docs


wtrocki:
When doing some restructure I found couple issues:
- Not all documents were linked in readme (we wanted to hide them)
- Service Discovery guide would be quite hard to split right now - it can be included into both but that will have kafka and registry content - we can do that as follow up

wtrocki:
>  product directory would contain both docs and code for each product, not just docs. 

That could be challenging to filter out some unneeded content etc.

Benefits of that:

- we can setup proper ownership structure for the repo (code for eng/docs
- automation doesn't need tons of ignore files (much simpler)

> Also, please move also the metrics-monitoring folder into the kafka docs folder (part of those docs).

Done

> And technically we need to move the _artifacts folder into the kafka folder too because those attributes are only for Kakfa, and then we need to add a new empty parent folder called _artifacts-common or something and use it for only attributes that apply to all products (currently we don't have anything like that yet).

Ok. I will add that to my list.

jbyrne-redhat:
> When doing some restructure I found couple issues:
> 
> * Not all documents were linked in readme (we wanted to hide them)
> * Service Discovery guide would be quite hard to split right now - it can be included into both but that will have kafka and registry content - we can do that as follow up

@wtrocki No need to worry about the service binding guide. I have a WIP PR (#389) that will make additions to that guide for Service Registry. _That_ can be the version we put it in the new Service Registry directory. So, no need to add a replica there in the meantime...

jbyrne-redhat:
@wtrocki While you're doing this, could you rename `service-discovery` to `service-binding`?

wtrocki:
@sterobin I think attributes require complete rework in separate PR due to conflicts that we might get when this is merged

wtrocki:
I will leave attributes for the content team to manage. 
Automation does work this way right now:

- _artifacts/document-attributes.adoc are injected as before - those should be global attributes. 
I do feel that including urls/anything like that should not be part of that file
- Attributes that are used in single files can be defined in the file or in the group folder (current structure) by using include.

Current structure promotes much better attributes higiene but I would like content team to reorganize them in separate PRs/as we go

CONTRIBUTING.adoc
```diff
@@ -324,6 +324,13 @@ NOTE: If you check in out of date attributes the build will fail
 . Having create the PR, automated tests will run. If they fail, use the error log to determine the problem, and fix it
 . Once your build is passing ask for review from a Subject Matter Expert (who will check for accuracy), a writer (who will check that the content is up to expected quality for substance, formatting, style, structure and consistency), and a developer who will ensure the steps covered by the quick start are the end to end test suite.
 
+== Adding new product to repository
+
+. Go to ./docs/product-mapping.yaml file and add new product to the list of products.
```
pmuir, line 329: I'm not sure we want to put the product mapping in the root of the docs directory like that. I wonder if it should be a dotfile?

docs/product-mapping.schema.json
```diff
@@ -0,0 +1,33 @@
+{
```
pmuir, line 1: I don't like this being in the root of the docs directory as it's really a tooling thing, it's not related to the content.

docs/product-mapping.yaml
```diff
@@ -0,0 +1,22 @@
+## File used in modular documentation generator to
```
pmuir, line 1: I don't like this being in the root of the docs directory as it's really a tooling thing, it's not related to the content.

pmuir (approved):
Looks good to me apart from I'd suggest putting the product-mapping.yaml perhaps as a doc file to make it clear it's part of the tooling for the repo, not the content.

CONTRIBUTING.adoc
wtrocki: `.product-mapping` sounds good (but yaml still preferred)

smccarthy-ie:
> I will leave attributes for the content team to manage. Automation does work this way right now:
> 
> * _artifacts/document-attributes.adoc are injected as before - those should be global attributes.
>   I do feel that including urls/anything like that should not be part of that file
> * Attributes that are used in single files can be defined in the file or in the group folder (current structure) by using include.
> 
> Current structure promotes much better attributes higiene but I would like content team to reorganize them in separate PRs/as we go

Cool, thanks @wtrocki Great that we can have a global attributes file like this. 

Those URLs are required for links to the Kafka or SR service, and for cross-reference links from one guide to another. These are common across the guides library for integration between different service content. It seems bad to duplicate these URLs in service-level attributes files?

wtrocki:
@smccarthy-ie I'm not the best person to comment on that. Attributes processing and generation stayed as it was before. I'm not going to change it and we can figure out some improvements for that in follow up PRs if needed.

sterobin:
@smccarthy-ie @wtrocki We'll need to discuss attributes on the CCS side and report back. It might be arguable whether we should in fact maintain the links and things globally or duplicated per product. I worry about the scalability of keeping global the URL links, because that would mean all services would need to maintain their URLs there, and the list could get pretty long, but maybe that's okay. At a minimum, we'll need to remove all the blatant prod-specific things, like `product-name` etc, so that each product can maintain their own prod attributes list in each prod dir. We'll see.

wtrocki:
Looks like all works end to end. Change is fully backwards compatible. Waiting for documentation ACK to we can merge it. 

Note: while I have made a lot of filesystem changes - they all are done without any change so other PRs **will not have any conflicts** after we merge. On the other hands there will be conflicts in this PR if we keep merging other docs changes

wtrocki:
@sterobin Can I get your approval on this? Happy to discuss attributes generation in follow up PRs. 

I will remove nested attributes that aren't used to not confuse anyone

pmuir (approved):

sterobin:
@wtrocki I was surprised to see this merged already. I hope it actually works with the pipeline to generate the same downstream results. I assumed we needed to test your branch out first before merging officially into main.

Regarding attributes, yes, let's chat with you and @pmuir . We discussed as a team and agreed that having just the one global attributes file divided up into prod sections and that's reused globally would be best, even if it's a large file. One problem is that with our current attribute injection approach for github rendering, that large attributes file content would then be at the top of every file in the repo as we do today, but just on a larger scale. Not an issue of process, just whether human contributors can tolerate that. If GitHub or adoc offers a way to truncate lines in a file, that would be ideal. Or if GitHub would finally just support includes, then we'd be set with a simple attributes include like we've always done for enterprise content.

Will circle back with you two about attributes and go from there.

wtrocki:
> @wtrocki I was surprised to see this merged already. I hope it actually works with the pipeline to generate the same downstream results. I assumed we needed to test your branch out first before merging officially into main.

We have actually tested this end to end on the PR level. Merge to main doesn't actually mean anything as we need to generate release. After release I see that results are the same. 


> If GitHub or adoc offers a way to truncate lines in a file, that would be ideal.

Including arguments is quite common practice/or transforming documents.

https://github.com/github/markup/issues/1095#issuecomment-1014386951

Adding upstream website to the repo is quite simple task so we might move to website instead of using github for rendering

PR #404, opened 2022-02-15
Updated some text in README.adoc to improve clarity and consistency.

AedinC:
Made minor changes using simpler and less formal language. Also moved sentences so they showed after code snippets instead of before them.

### DIFF
--- a/docs/kafka/kafka-bin-scripts-kafka/README.adoc
+++ b/docs/kafka/kafka-bin-scripts-kafka/README.adoc
@@ -65,7 +65,7 @@ NOTE: The Kafka scripts are part of the open source community version of Apache 
 endif::[]
 
 When you download and extract the Apache Kafka distribution, the `bin/` directory (or the `bin\windows\` directory if you're using Windows) of the distribution contains a set of shell scripts that enable you to interact with your Kafka instance.
-With the scripts, you can produce and consume messages, and perform various operations against the Kafka APIs to administer topics, consumer groups, and other resources.
+With the scripts, you can produce and consume messages using your Kafka instances. You can also perform various other operations against the Kafka APIs to administer topics, consumer groups, and other resources.
 
 NOTE: The command examples in this quick start demonstrate how to use the Kafka scripts on Linux and macOS. If you're using Windows, use the Windows versions of the scripts. For example, instead of the `__<Kafka-distribution-dir>__/bin/kafka-console-producer.sh` script, use the `__<Kafka-distribution-dir>__\bin\windows\kafka-console-producer.bat` script.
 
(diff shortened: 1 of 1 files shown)


docs/kafka/kafka-bin-scripts/README.adoc
```diff
@@ -141,44 +141,48 @@ You can use the `kafka-console-producer` script to produce messages to Kafka top
...
 +
 --
-This example uses the `kafka-topics` script to create the `my-other-topic` Kafka topic with the default settings.
 
 ifdef::qs[]
 The `<bootstrap_server>` is the bootstrap server endpoint for your Kafka instance. You copied this information previously for the Kafka instance in {product} by selecting the options menu (three vertical dots) and clicking *Connection*.
 endif::[]
 
-.Using the `kafka-topics` script to create a Kafka topic
+Using the `kafka-topics` script to create a Kafka topic
```
bhardesty, line 159: This might be accidental - I think you still want the preceding period so that the text is styled as a title for the code block.
```suggestion
.Using the `kafka-topics` script to create a Kafka topic
```

docs/kafka/kafka-bin-scripts/README.adoc
```diff
@@ -141,44 +141,48 @@ You can use the `kafka-console-producer` script to produce messages to Kafka top
...
++
+The preceding example uses the `kafka-topics` script to create the `my-other-topic` Kafka topic with the default settings.
 
 . Enter the following command to start the `kafka-console-producer` script.
 +
 --
-This example uses the SASL/PLAIN authentication mechanism with the credentials that you saved in the `{property-file-name}` file. This example produces messages to the `my-other-topic` example topic that you created.
 
-.Starting the `kafka-console-producer` script
+Starting the `kafka-console-producer` script
```
bhardesty, line 173: Same comment as before. The period just styles the text as a title.

docs/kafka/kafka-bin-scripts/README.adoc
```diff
@@ -207,13 +211,13 @@ You can use the `kafka-console-consumer` script to consume messages from Kafka t
 * You used the `kafka-console-producer` script to produce example messages to a topic.
 
 .Procedure
+. Open a second terminal window or tab, separate from the producer.
 
-. On the command line in a separate terminal from your producer, enter the following command to start the `kafka-console-consumer` script.
+. On the command line, enter the following command to start the `kafka-console-consumer` script.
```
bhardesty, line 216: I would keep the phrase about using a separate terminal. The reason is that while the producer is running, you can't run the consumer in the same terminal session. Having the producer and consumer running side-by-side is helpful for seeing the messages go from the producer to the consumer.

bhardesty (commented):
Looks great, @AedinC! I just had a couple minor suggestions.

docs/kafka/kafka-bin-scripts/README.adoc
AedinC: @bhardesty I added a step before that reads 'Open a second terminal window or tab, separate from the producer.' so that information is still there.

PR #405, opened 2022-02-15
fix: remove service-binding guide from RHOSR

pmuir:


smccarthy-ie:
LGTM, thanks! 

@jbyrne-redhat Does `service-binding` still need be a top-level dir for RHOSAK for now (i.e., moving to `kafka/service-binding` and `registry/service-binding` later)?

smccarthy-ie (approved):

pmuir:
@jbyrne-redhat can we do it as a separate PR just to avoid overloading this one?

jbyrne-redhat:
> @jbyrne-redhat can we do it as a separate PR just to avoid overloading this one?

Sure, @pmuir - no problem.

PR #407, opened 2022-02-16
Rework command output in CLI guides

bhardesty:
@jbyrne-redhat @bredamc @wtrocki This is the former PR 399, with all of your review comments incorporated.

### DIFF
--- a/docs/kafka/rhoas-cli-getting-started-kafka/README.adoc
+++ b/docs/kafka/rhoas-cli-getting-started-kafka/README.adoc
@@ -98,20 +98,6 @@ This example creates a Kafka instance called `my-kafka`.
 [source,shell]
 ----
 $ rhoas kafka create --name my-kafka
-{
-  "cloud_provider": "aws",
-  "created_at": "2021-03-17T18:33:50.000799005Z",
-  "href": "/api/managed-services-api/v1/kafkas/1ptcsACdx7HVzacKZBA6HRRa1oW",
-  "id": "1ptcsACdx7HVzacKZBA6HRRa1oW",
-  "kind": "Kafka",
-  "multi_az": true,
-  "name": "my-kafka",
-  "owner": "developer",
-  "region": "us-east-1",
-  "status": "accepted",
-  "updated_at": "2021-03-17T18:33:50.000799005Z"
-}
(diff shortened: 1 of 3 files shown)


docs/registry/rhoas-cli-service-registry/README.adoc
```diff
@@ -380,45 +280,38 @@ After you register a {registry} artifact, you can download the artifact to updat
...
+* You've created a {registry} instance with at least one artifact.
 
 .Procedure
 
-Download the artifact in either of the following ways:
-
-* Use the `rhoas service-registry artifact get` command and specify the artifact ID.
-* Use the `rhoas service-registry artifact download` command and specify the global ID.
-
+* Download the artifact by specifying either the artifact ID or the global ID.
```
bredamc, line 287: In the original text, I used a bullet list to highlight that you use a different command depending on which ID you use ("get" for artifact ID, "download" for global ID). I don't think that distinction is clear in this rewrite?

jbyrne-redhat (approved):
LGTM, @bhardesty.

docs/registry/rhoas-cli-service-registry/README.adoc
bhardesty: I tried to bring that procedure more in line with the modular documentation standards, where a single step is represented by a bullet point. This particular procedure is a single step because there's only one action to take (download the artifact), but there're two options for completing it. The other way to handle this would be to use sub-bullets, which seems to be more complicated to follow:

* Download the artifact by doing one of the following:
    * Download the artifact by artifact ID...
    * Download the artifact by global ID...

Does that seem to flow better? If so, I'm happy to change it.

bredamc: Thanks, @bhardesty -- my apologies, I'd forgotten that a single step had to be a bullet point. In that case, I'd prefer to keep my original text, and just apply the bullet point and the sub-bullet points to that text? I prefer this approach to avoid the repetition of "Download the artifact by", and make it easier to see that different commands are used (and also to ensure that "following" is used as an adjective before a noun). Have I convinced you, or would you still prefer the one-line version? :)

bhardesty: No problem - I made that update.

PR #409, opened 2022-02-23
Fix access instruction

sterobin:
Fixes a bug I introduced from a misunderstanding of functionality.


jbyrne-redhat (approved):
Updates look good, @sterobin  - no changes to request.

PR #418, opened 2022-03-08
Reformatted the links in prereqs

ritz303:
After personal review, I noticed the prerequisite link format was only working for the [preview on the tile](https://drive.google.com/file/d/16VStgvWl71nJ_QcmGu2MfypEKMfv5Cqa/view?usp=sharing), but [broken](https://drive.google.com/file/d/16-j8jTrPK1QR4FNSOaXDZpnT3G5Oh4EM/view?usp=sharing) at the start of the quick start guide.  I've updated the link syntax so the links should work in both places.


wtrocki (approved):

ritz303:
@pmuir @wtrocki : I double checked what the OpenShift Service Registry quick start guide was doing with their links in the prereqs, and it looks like that syntax (<a href=http://example.com>title</a>) doesn't seem to work either, because it's broken at the beginning of their quick start guide, but works in the preview of the prereqs on the tile.  I would like to keep the links in the prereqs.  Any guidance on the proper syntax here, so, the links work in both places?  I think I would like to put a hold on merging this PR until I have this figure out.  I don't consider this a blocker for the release.

jschuler:
We have 2 related issues open 
https://github.com/patternfly/patternfly-quickstarts/issues/142
https://github.com/patternfly/patternfly-quickstarts/issues/138
I'll see if I can get a new build out with a fix for rendering links in the introduction prereqs

PR #411, opened 2022-02-25
fix: swap JDK to recommend OpenJDK instead

wtrocki:
Replace OracleJDK recommendation with OpenJDK (17). 
Actual description can be reviewed and improved depending on how we want to put recommendation in place.

PR #419, opened 2022-03-10
Jk 7743 - Removed JDK Prerequisite

AedinC:
As per instruction from QE (https://issues.redhat.com/browse/MGDSTRM-7743), I removed the prerequisite to have JDK installed.


wtrocki:
@AedinC We already have this PR: https://github.com/redhat-developer/app-services-guides/pull/411

Since you are working on it. Do you mind applying suggestions from @sterobin

wtrocki:
What we looking for is


```
* https://adoptopenjdk.net/[JDK^] 11 or later is installed. (The latest LTS version of OpenJDK is recommended.)
```

PR #428, opened 2022-03-18
Update .product-mapping.yml

bhardesty:
Setting the directory for OpenShift Connectors. Doc work should begin soon.


wtrocki:
To commit empty folder we can add .keep file to it that will be removed later.

wtrocki (approved):

PR #449, opened 2022-04-11
Updated the procedure for Reviewing consumer group properties

fmcdonal:


fmcdonal:
@mmurphy@redhat.com, please review

wtrocki:
FYI @mmurphy

jbyrne-redhat:
@fmcdonal Could you generate an HTML preview and provide a link here? Thanks!

fmcdonal:
http://file.rdu.redhat.com/~fmcdonal/MGDSTRM-7853/MGDSTRM-7853_README_140422.html generated - please review
@mmurphy and @jbyrne-redhat

jbyrne-redhat:
Thanks for generating the HTML preview @fmcdonal - that helps. 

Now that I see the output, I don't think the structure of the procedure is right. Currently, the paragraph beginning "Alternatively..." appears as an alternative set of steps for just step 3 of the main procedure. Which they aren't - they are an alternative to steps 1, 2, **and** 3.

I'm not sure we need to add a whole alternative procedure anyway. Instead, I think we could do something like the following:

```
.Procedure
. In the {product-kafka} web console, go to *Streams for Apache Kafka* > *Kafka Instances* and click the name of the Kafka instance that contains the consumer groups that you want to review.
. Select the *Consumer groups* page.
. Select the options icon (three vertical dots) for the relevant consumer group and click *View partition offsets*.

NOTE: You can also view consumer groups for a specific topic. In your Kafka instance, select the *Topics* page and then click the name of a topic. Select the *Consumer groups* page. For a given consumer group, select the options icon (three vertical dots) and click *View partition offsets*.
```

jbyrne-redhat (approved):

PR #467, opened 2022-04-21
Add purchasing additional resource to GS quick start

bhardesty:
Added an additional resource to the Kafka GS quick start to point users to the Kafka Overview page that contains the info about how to purchase a subscription. 

As an additional resource, this won't show up in the quick start itself (the additional resources are only included in the customer portal version). I don't see a great way to add the pointer to the purchasing info to the quick start itself. The conclusion would probably be the best place, but it's already got a fair amount going on.

### DIFF
--- a/docs/kafka/getting-started-kafka/README.adoc
+++ b/docs/kafka/getting-started-kafka/README.adoc
@@ -307,6 +307,7 @@ endif::[]
 
 [role="_additional-resources"]
 == Additional resources
+* https://console.redhat.com/application-services/streams/overview[Purchase a subscription to {product-long-kafka}]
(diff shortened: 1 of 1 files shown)


jbyrne-redhat:
> Added an additional resource to the Kafka GS quick start to point users to the Kafka Overview page that contains the info about how to purchase a subscription.
> 
> As an additional resource, this won't show up in the quick start itself (the additional resources are only included in the customer portal version). I don't see a great way to add the pointer to the purchasing info to the quick start itself. The conclusion would probably be the best place, but it's already got a fair amount going on.

@bhardesty For the console quick start, users will need a Kafka instance. I'm not exactly sure when we'll stop offering trial instances, but even if it's for some time to come, I'd like to believe we'll be putting the purchasing info prominently in front of them in the console before they create one! It makes perfect sense as an additional resource in the Customer Portal version, however.

jbyrne-redhat (approved):
LGTM!

PR #478, opened 2022-04-25
Use token endpoint apicurio serdes

carlesarnal:
This PR addresses two things:

- Use token endpoint instead of auth URL + realm.
- Update Quarkus version used in example.


wtrocki:
Reviewing

wtrocki:
Works end to end but I will merge it with our docs change as well

wtrocki (approved):

rkpattnaik780 (approved):

PR #479, opened 2022-04-25
Fix conclusion in getting started quick start

bhardesty:
The conclusion in the Getting Started with Kafka quick start has two paragraphs, and the second paragraph of text was not displayed. I added a delimited block to fix the issue.


docs/kafka/getting-started-kafka/README.adoc
```diff
@@ -317,9 +317,11 @@ endif::[]
 
 ifdef::qs[]
 [#conclusion]
+====
 Congratulations! You successfully completed the {product-kafka} Getting Started quick start, and are now ready to use the service.
 
 You can use Kafka scripts to check that you can connect with your Kafka instance.
+====
```
jbyrne-redhat, line 324: Looks good, @bhardesty.

My only suggestion is that if we're calling out Kafka scripts specifically here, then we modify `quickstart.yml` to make Kafka scripts the first of the "next quick starts" that we suggest.

Alternatively, we modify the last line of the conclusion to mention both Kafka and Kafkacat: "You can use either Kafka scripts or Kakfacat to check....".

I think the second option is my preference.

jbyrne-redhat (changes requested):
One further update I think we should make, @bhardesty.

docs/kafka/getting-started-kafka/README.adoc
bhardesty: @jbyrne-redhat Yes, I see what you mean. I went with the second suggestion.

jbyrne-redhat (approved):

PR #484, opened 2022-04-28
docs: change cli guides with updated command names

rkpattnaik780:
Documentation has been updated as certain context sub-commands were renamed.

`rhoas context use-kafka` => `rhoas context set-kafka`
`rhoas context use-service-registry` => `rhoas context set-service-registry`

For [app-services-cli/1542](https://github.com/redhat-developer/app-services-cli/pull/1542)

### DIFF
--- a/docs/kafka/rhoas-cli-getting-started-kafka/README.adoc
+++ b/docs/kafka/rhoas-cli-getting-started-kafka/README.adoc
@@ -130,7 +130,7 @@ $ rhoas context status kafka
 [NOTE]
 ====
 If you have multiple Kafka instances,
-you can make the current context switch to a different instance by using the `rhoas context use-kafka` command.
+you can make the current context switch to a different instance by using the `rhoas context set-kafka` command.
 ====
 --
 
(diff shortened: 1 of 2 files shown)


wtrocki:
@sterobin Do you know if we have some form of rapid update of the downstream docs for such PRs? Is there capability to deploy docs on demand?

wtrocki (approved):
Good to update. I think we can merge and then try to time release of the CLI when content updates

rkpattnaik780:
> Good to update. I think we can merge and then try to time release of the CLI when content updates

Was thinking if we should use `aliases` more for scenarios like these.

wtrocki:
Aliases would work but might be even more confusing :D and eventually we will need to introduce breaking change. best to do it faster than later.

sterobin:
@wtrocki Yes, if there's a critical bug or something that must be updated asap before the next release/publish cycle, just notify one of us on the docs team and we'll create the release upstream and republish just the single doc in question.

wtrocki:
This changes have been release week ago. We have been seeing users using wrong commands in the logs. 
Merging this to make for the release tommorow.

PR #495, opened 2022-04-29
Fix all QS tag blocks.

sterobin:


bhardesty (approved):

wtrocki:
We could release it :)

PR #501, opened 2022-05-03
Update URL for core registry REST API docs to latest upstream version

smccarthy-ie:
Simple fix for latest 2.2.x version of upstream REST API docs.


bredamc (approved):
LGTM :)

PR #514, opened 2022-05-23
Add note to client guides about message browsing in console

jbyrne-redhat:
**Updates**
  
 - Add a note to guides that describe connecting clients (that is, Kcat, Kafka scripts, Quarkus, Node.js, service binding) to inform users that they can also browse messages in the console
 - Add {message-browsing-url-kafka} attribute for message browsing guide
 - Update all guides with new attribute


sterobin (commented):
@jbyrne-redhat Makes sense to me. I can't really tell without the doc previews if the notes are rendering properly in the output, but if you've checked for certain and all looks good, I'll take your word for it rather than make you upload multiple docs previews. :)

One question: I know I raised this before, but as of our style today, all of the refs to doc should be italicized. I know we're iterating around things like this at the moment with our style discussions, but the majority of the trend established initially was to italicize doc titles in the customer portal. This may be worth revisiting since we don't italicize any other docs outside of Red Hat, which we seem to consider more web pages that actual docs, whereas ours we seem to consider more docs than pages.

Wdyt for now?

sterobin (approved):
@jbyrne-redhat Agreed about the ital probably not being the way to go, but best to do for now for consistency. Looking forward to that discussion. Approved. Thanks.

docs/kafka/service-binding-kafka/README.adoc
```diff
@@ -752,3 +755,5 @@ You should now see that the producer has connected to the Kafka instance. The pr
 . Open the logs for the pod of the consumer component, in the same way that you did for the producer component.
 +
 You should now see that the consumer has connected to the Kafka instance. The consumer displays the same country names that the producer sends to the `countries` Kafka topic, and in the same order.
++
```
wtrocki, line 758: Should we let people know at the end or begining ?

wtrocki (approved):

docs/kafka/service-binding-kafka/README.adoc
jbyrne-redhat: It's a good question, @wtrocki. But I think we're putting this information in the appropriate place. It's not until the end of each example that users actually have messages to see.

wtrocki: Yes. Although produce UI will be also added in upcomming sprints.

PR #515, opened 2022-05-23
fix: add metrics token support

wtrocki:
Metrics guide should follow the same pattern to tell users to copy token url from UI and CLI. 
Alternatively we can give them curl to obtain that value (if somehow they did not copied it before)

Source: https://issues.redhat.com/browse/MGDSTRM-8460

### DIFF
--- a/docs/kafka/metrics-monitoring-kafka/README.adoc
+++ b/docs/kafka/metrics-monitoring-kafka/README.adoc
@@ -165,7 +165,7 @@ This procedure follows the https://prometheus.io/docs/prometheus/latest/configur
 * You've installed a Prometheus instance in your monitoring environment. For installation instructions, see https://prometheus.io/docs/prometheus/latest/getting_started/[Getting Started] in the Prometheus documentation.
 
 .Procedure
-. In your Prometheus configuration file, add the following information. Replace `<kafka_instance_id>` with the ID of the Kafka instance. Replace `<client_id>` and `<client_secret>` with the generated credentials for your service account.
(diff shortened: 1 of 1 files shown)


sterobin (commented):
@wtrocki Excellent catch on this! I pushed a commit with some additional tweaks. Hope that's okay, was just easier.

- I revised the text a bit to use the same approach and language as we do with the same info in other docs, for consistency.
- Added the prereq to also be consistent with how we address it in other docs and to also make it clearer.
- Also made the same change in the similar code example that follows it.
- Removed the bit about using curl to obtain the token ID. We already direct them to the UI for everything else they need to obtain, so for simplicity I think it's better IMO to stick with that approach and not introduce additional options for just that one thing. I tried squeezing it in as a note, but just seemed out of place and unnecessary. Open to discussion of course if you disagree.

@AedinC ^^

wtrocki:
Amazing collaboration. This was low effort PR from my side. Now it is very well ready to be merged. Thank you so much for making the change.

Once that is merged all token urls will be consistently provided by UI and CLI. Some example outputs still have it but that is not affecting users in any way

sterobin:
@wtrocki ack

@AedinC Since the related Jira belongs to you, can you have a look and approve if you agree? Just to ensure I'm not stepping on your toes. Once you're good, go ahead with the merge. Thanks.

AedinC (approved):
LGTM! Thanks @sterobin!

wtrocki:
LGTM

PR #517, opened 2022-05-24
Updated streaming unit size info and new screenshot

AedinC:
Updated the following:
-Prerequisite with info about streaming unit and link to service limits article
-Added new instance steps
-Changed configuration image


docs/kafka/getting-started-kafka/README.adoc
```diff
@@ -121,6 +123,7 @@ endif::[]
 * *Cloud region*: Select `US East, N. Virginia`.
 * *Availability zones*: Select `Multi`.
 ////
+. Use the slider to select the required instance size. Instances with more streaming units have more capacity. In {product-kafka}, you can create Kafka instances of 1 or 2 streaming units, however, instances larger than 1 streaming unit are currently available only as a Technology Preview and should not be used in production.
```
jbyrne-redhat, line 126: A couple of suggestions:

I think I'd eliminate any notion of a _required_ size here. Instead, I think you can simply say "Use the slider to select an instance size. Instances with more streaming units have more capacity."

docs/kafka/getting-started-kafka/README.adoc
```diff
@@ -107,11 +107,13 @@ Use the {product-kafka} web console to create and configure a Kafka instance for
 ifndef::qs[]
 .Prerequisites
 * You're logged in to the {product-kafka} web console at {service-url-kafka}[^].
+* You're familiar with the different sized Kafka instances that you can create in {product-kafka}, and the capacities that correspond to those sizes. For more information about the capacities and service limits of your Kafka instance, see https://access.redhat.com/articles/5979061[Red Hat OpenShift Streams for Apache Kafka Service Limits].
```
jbyrne-redhat, line 110: Because it is a compound adjective, I suggest "different-sized".

jgiardino (commented):
These updates look good! 

This is my first time dealing with this scenario, but do our quick start guides always assume a user has a subscription once the service is GA?

Should there be any comment about this guide assumes that you have a subscription?

I don't see this question as a blocker for this PR, but it would be good to understand what our overall strategy is so that we can communicate that to other services that might face similar scenarios.

AedinC:
> These updates look good!
> 
> This is my first time dealing with this scenario, but do our quick start guides always assume a user has a subscription once the service is GA?
> 
> Should there be any comment about this guide assumes that you have a subscription?
> 
> I don't see this question as a blocker for this PR, but it would be good to understand what our overall strategy is so that we can communicate that to other services that might face similar scenarios.

@jgiardino Great question. There will be another update made soon to this guide regarding trial sizes. I can definitely add a comment here to make it clearer these updated steps are for when the user has a subscription.

docs/kafka/getting-started-kafka/README.adoc
```diff
@@ -123,9 +123,14 @@ endif::[]
 * *Cloud region*: Select `US East, N. Virginia`.
 * *Availability zones*: Select `Multi`.
 ////
-. Use the slider to select the required instance size. Instances with more streaming units have more capacity. In {product-kafka}, you can create Kafka instances of 1 or 2 streaming units, however, instances larger than 1 streaming unit are currently available only as a Technology Preview and should not be used in production.
-. Click *Create instance* to start the creation process for your Kafka instance.
+. If you have purchased a subscription for {product-kafka}, you can use the slider to select an instance size. Instances with more streaming units have more capacity.
```
jbyrne-redhat, line 126: Two suggestions:

 - "If you've purchased.." (stay consistent with informal voice elsewhere in this guide)
 - "...use the slider" (eliminate "you can")

docs/kafka/getting-started-kafka/README.adoc
```diff
@@ -123,9 +123,14 @@ endif::[]
...
 +
+NOTE: In {product-kafka}, you can create Kafka instances of 1 or 2 streaming units, however, instances larger than 1 streaming unit are currently available only as a Technology Preview and should not be used in production.
```
jbyrne-redhat, line 128: I think it _might_ improve readability to split this into two sentences, the second starting with "However,..."

I don't think this would be a big deal if it was in a standard paragraph. But this note is quite an important one, and should be very easy to parse.

docs/kafka/getting-started-kafka/README.adoc
jbyrne-redhat, line 126: Use a contraction here to stay consistent  with the informal voice elsewhere in the guide. "If you've purchased..."

docs/kafka/getting-started-kafka/README.adoc
jbyrne-redhat: I wonder if we should consider an additional sentence to the note, along the lines of the following: "If you're creating a trial instance, the instance size is preconfigured."

AedinC: @jbyrne-redhat Yes. I can add that. I was thinking about adding something about trial instances but wasn't sure if this straying into @fmcdonal update on trial instances.

jbyrne-redhat: @AedinC Well, maybe decide what is best between you! It just struck me that users might be confused about the absence of the slider if creating a trial instance.

jbyrne-redhat (approved):
These updates look good to me, @AedinC  - nice job.

PR #518, opened 2022-05-24
Added information about trial and production instance limits

AedinC:
New trial and production instance limit values added.

### DIFF
--- a/docs/kafka/metrics-monitoring-kafka/README.adoc
+++ b/docs/kafka/metrics-monitoring-kafka/README.adoc
@@ -143,8 +143,7 @@ The Kafka instance type determines the maximum storage in the broker, as shown i
 
 * `kafka_broker_quota_totalstorageusedbytes`: Amount of storage, in bytes, that is currently used by partitions in the broker. The storage usage depends on the number and retention configurations of the partitions. This metric must stay below the `kafka_broker_quota_softlimitbytes` metric setting.
 
-* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions in the cluster. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`. This count includes partitions from internal Kafka topics, such as `\__consumer_offsets` and `__transaction_state`.
-This metric is similar to the `kafka_topic:kafka_topic_partitions:sum` cluster metric.
+* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions in the cluster. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`. This count includes partitions from internal Kafka topics, such as `\__consumer_offsets` and `__transaction_state`. This metric is similar to the `kafka_topic:kafka_topic_partitions:sum` cluster metric.
 +
 The Kafka instance type determines the maximum storage in the broker, as shown in the following values:
 
--- a/docs/kafka/metrics-monitoring-kafka/README.adoc
+++ b/docs/kafka/metrics-monitoring-kafka/README.adoc
@@ -86,34 +86,44 @@ As a developer or administrator, you can view metrics in {product-kafka} to visu
 Cluster metrics::
 +
 --
-* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: Number of incoming bytes per second for the cluster in the last five minutes. This metric represents all the data that producers are sending to topics in the cluster. +
-Your chosen Kafka instance type determines the maximum incoming byte rate:
+* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: Number of incoming bytes per second for the cluster in the last five minutes. This metric represents all the data that producers are sending to topics in the cluster.
++
+The Kafka instance type determines the maximum incoming byte rate, as shown in the following values:
+
 ** 1 MiB per second for a trial instance
 ** 50 MiB per second for a production instance with a size of 1 streaming unit
 ** 100 MiB per second for a production instance with a size of 2 streaming units
 
-* `kafka_namespace:haproxy_server_bytes_out_total:rate5m`: Number of outgoing bytes per second for the cluster in the last five minutes. This metric represents all the data that producers are receiving from topics in the cluster.  +
-Your chosen Kafka instance type determines the maximum outgoing byte rate:
+* `kafka_namespace:haproxy_server_bytes_out_total:rate5m`: Number of outgoing bytes per second for the cluster in the last five minutes. This metric represents all the data that producers are receiving from topics in the cluster.
++
+The Kafka instance type determines the maximum outgoing byte rate, as shown in the following values:
+
 ** 1 MiB per second for a trial instance
 ** 100 MiB per second for a production instance with a size of 1 streaming unit
 ** 200 MiB per second for a production instance with a size of 2 streaming units
 
-* `kafka_namespace:kafka_server_socket_server_metrics_connection_count:sum`: Number of current client connections to the cluster. Kafka clients use persistent connections to interact with brokers in the cluster. For example, a consumer holds a connection to each broker it is receiving data from and a connection to its group coordinator. +
-Your chosen Kafka instance type determines the maximum number of active connections:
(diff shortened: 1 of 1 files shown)


AedinC:
> @AedinC Do you mind adding the doc preview link? Easier for review and sanity check. Thanks.

Sorry- it's [http://file.ams.redhat.com/aecollin/JK-8598/README.html](http://file.ams.redhat.com/aecollin/JK-8598/README.html)

docs/kafka/metrics-monitoring-kafka/README.adoc
```diff
@@ -86,27 +86,57 @@ As a developer or administrator, you can view metrics in {product-kafka} to visu
...
-
-* `kafka_namespace:kafka_server_socket_server_metrics_connection_creation_rate:sum`: Number of client connection creations per second for the cluster. Kafka clients use persistent connections to interact with brokers in the cluster. A constant high number of connection creations might indicate a client issue. The maximum connection creation rate is 100 per second.
-
-* `kafka_topic:kafka_topic_partitions:count`: Number of topics in the cluster. This does not include internal Kafka topics, such as `\__consumer_offsets` and `__transaction_state`. The maximum number of topics is 1000.
-
-* `kafka_topic:kafka_topic_partitions:sum`: Number of partitions across all topics in the cluster. This does not include partitions from internal Kafka topics, such as `\__consumer_offsets` and `__transaction_state`. The maximum number of partitions is 1000.
+* `kafka_namespace:haproxy_server_bytes_in_total:rate5m`: Number of incoming bytes per second for the cluster in the last five minutes. This metric represents all the data that producers are sending to topics in the cluster. +
+Your chosen Kafka instance type determines the maximum incoming byte rate:
```
sterobin, line 90: Suggest the following rewording here and likewise in all other entries. Note the addition too of "as shown ...." The only reason for this is that in our IBM style discussion we covered that IBM prefers to concretely introduce lists and such, even though the format is less streamlined. Also note that you need a line space here and in all the other entries, which I've also shown in my suggestion below.

```
... in the cluster.
+
The Kafka instance type determines the maximum incoming byte rate, as shown in the following values:
```

docs/kafka/metrics-monitoring-kafka/README.adoc
```diff
@@ -86,27 +86,57 @@ As a developer or administrator, you can view metrics in {product-kafka} to visu
...
 
-* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions in the cluster. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`. This count includes partitions from internal Kafka topics, such as `\__consumer_offsets` and `__transaction_state`. The maximum number of partitions is 1000. This metric is similar to the `kafka_topic:kafka_topic_partitions:sum` cluster metric.
+* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions in the cluster. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`. This count includes partitions from internal Kafka topics, such as `\__consumer_offsets` and `__transaction_state`. +
+Your chosen Kafka instance type determines the maximum storage in the broker:
+** 100 for a trial instance
+** 1500 for a production instance with a size of 1 streaming unit
+** 3000 for a production instance with a size of 2 streaming units  +
+This metric is similar to the `kafka_topic:kafka_topic_partitions:sum` cluster metric.
```
sterobin, line 139: I suggest keeping this line as part of the metric description before the instance type stuff, where it was before. It makes more sense to me as part of the description itself. It's also more consistent with how we do the same with another metric where we mention something similar. If you were to keep it, you'd still need to add a line space before it in the format I mentioned in my other comment, but best imo to run it with the description.

sterobin (commented):
@AedinC Thanks for this! Just a couple comments.

docs/kafka/metrics-monitoring-kafka/README.adoc
```diff
@@ -86,27 +86,72 @@ As a developer or administrator, you can view metrics in {product-kafka} to visu
...
+
 
 * `kafka_broker_quota_totalstorageusedbytes`: Amount of storage, in bytes, that is currently used by partitions in the broker. The storage usage depends on the number and retention configurations of the partitions. This metric must stay below the `kafka_broker_quota_softlimitbytes` metric setting.
 
-* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions in the cluster. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`. This count includes partitions from internal Kafka topics, such as `\__consumer_offsets` and `__transaction_state`. The maximum number of partitions is 1000. This metric is similar to the `kafka_topic:kafka_topic_partitions:sum` cluster metric.
+* `kafka_controller_kafkacontroller_global_partition_count`: Number of partitions in the cluster. Only the broker that is the current controller in the cluster reports this metric. Any other brokers report `0`. This count includes partitions from internal Kafka topics, such as `\__consumer_offsets` and `__transaction_state`.
+This metric is similar to the `kafka_topic:kafka_topic_partitions:sum` cluster metric.
```
sterobin, line 147: @AedinC I'd just bring it up to follow in the previous paragraph, no line break. The non-spaced break of a line can sometimes render it all in a paragraph but sometimes not, I've found. So you can keep it here if you like the "every line a sentence" approach in the source, but that's not our norm so far. Just have a look and make sure it looks good in the rendering. Thanks.

sterobin (approved):

docs/kafka/metrics-monitoring-kafka/README.adoc
AedinC: Thanks @sterobin. I'll change that.

docs/kafka/metrics-monitoring-kafka/README.adoc
```diff
@@ -86,27 +86,71 @@ As a developer or administrator, you can view metrics in {product-kafka} to visu
...
+
+** 1 MiB per second for a trial instance
+** 50 MiB per second for a production instance with a size of 1 streaming unit
+** 100 MiB per second for a production instance with a size of 2 streaming units
+
+* `kafka_namespace:haproxy_server_bytes_out_total:rate5m`: Number of outgoing bytes per second for the cluster in the last five minutes. This metric represents all the data that producers are receiving from topics in the cluster.
++
+The Kafka instance type determines the maximum outgoing byte rate, as shown in the following values:
+
+** 1 MiB per second for a trial instance
```
wtrocki, line 101: [Optional] I'm wondering if we want to duplicate that information across (it increases likelyhood of forgetting about changing those). We have limits presented in tabular form here: https://access.redhat.com/articles/5979061

I think it will be better for our business capabilities if we can introduce new sizes in one place and point users to the one size reference article that we always keep updating.

wtrocki:
Looks good.

docs/kafka/metrics-monitoring-kafka/README.adoc
AedinC: > [Optional] I'm wondering if we want to duplicate that information across (it increases likelyhood of forgetting about changing those). We have limits presented in tabular form here: https://access.redhat.com/articles/5979061
> 
> I think it will be better for our business capabilities if we can introduce new sizes in one place and point users to the one size reference article that we always keep updating.

Thanks @wtrocki I think we still need to update limit amounts here.

wtrocki: Do not see this comment as comment on your PR. This is more on strategy we should pick for documenting values like that. I always recommend to keep those in one place. 

This values are very meaningful for users that would likely to make purchase decisions based on them. Having them out of date or updated in different timing has business implications etc. I'm ok to get those merged into that doc but if we do that we need to have registry for all documents that mention sizes to update them every time something changes (and things would change for sure)

AedinC: > Do not see this comment as comment on your PR. This is more on strategy we should pick for documenting values like that. I always recommend to keep those in one place.
> 
> This values are very meaningful for users that would likely to make purchase decisions based on them. Having them out of date or updated in different timing has business implications etc. I'm ok to get those merged into that doc but if we do that we need to have registry for all documents that mention sizes to update them every time something changes (and things would change for sure)

Yes, agree it causes more maintenance for us.

PR #519, opened 2022-05-24
fix: initial version of whats new file

wtrocki:
Idea for what is new is that every large change or feature that we would add to guides needs to be at least mentioned in the whats new section.

This could be long living document similar to release notes but much more sophisticated and curated by eng and content


wtrocki:
FYI @jbyrne-redhat . Empty file for now but I can add some initial content

PR #523, opened 2022-06-02
Incorrect environment variables and env file name

jackdelahunt:
Just going through some of the guides node with Kafka and there were some issues.

The example code used in the example [here](https://github.com/nodeshift-starters/reactive-example) uses different environment variables for consumer and producer backends. The specific code examples are below.

- [Consumer](https://github.com/nodeshift-starters/reactive-example/blob/kafkajs/consumer-backend/consumer.js#L26-L32)
- [Producer](https://github.com/nodeshift-starters/reactive-example/blob/kafkajs/producer-backend/producer.js#L17-L23)

Also in the docs it mentions creating a file called `rhoas.env` this did not set the environment variables even in root or the working directory. Changing this to `.env` worked.

### DIFF
--- a/docs/kafka/nodejs-kafka/README.adoc
+++ b/docs/kafka/nodejs-kafka/README.adoc
@@ -135,7 +135,7 @@ To enable your Node.js application to access a Kafka instance, you must configur
 * The generated credentials for your {product-kafka} service account
 * The Simple Authentication and Security Layer (SASL) mechanism that the client will use to authenticate with the Kafka instance
 
-In this task, you'll create a new configuration file called `rhoas.env`. In this file, you'll set the required bootstrap server and client credentials as environment variables.
+In this task, you'll create a new configuration file called `.env`. In this file, you'll set the required bootstrap server and client credentials as environment variables.
 
 .Prerequisites
 ifndef::qs[]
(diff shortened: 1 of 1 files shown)


docs/kafka/nodejs-kafka/README.adoc
```diff
@@ -146,16 +146,16 @@ endif::[]
...
+. In your IDE, create a new file. Save the file with the name `.env`, at the root level of the `reactive-example` directory for the cloned repository.
 
-. In the `rhoas.env` file, add the lines shown in the example. These lines set the bootstrap server and client credentials as environment variables to be used by the Node.js application.
+. In the `.env` file, add the lines shown in the example. These lines set the bootstrap server and client credentials as environment variables to be used by the Node.js application.
 +
-.Setting environment variables in the rhoas.env file
+.Setting environment variables in the .env file
 [source,subs="+quotes"]
 ----
-KAFKA_HOST=__<bootstrap_server>__
```
wtrocki, line 156: Those values will come from `rhoas cli` context. Wondering why we needed to change them?

@rkpattnaik780 did you made any changes on node.js side to support new environment variables?

rkpattnaik780: Just realized the changes hadn't been pushed. Will be raising the PR now.

The values are not obtained from the CLI, this guide is around UI. The environment variables were changed according to the current configuration file format.

rkpattnaik780:
NIce catch @jackdelahunt :100: 
Thanks a ton! 

I have raised a [PR](https://github.com/nodeshift-starters/reactive-example/pull/72) for the changes in the example mentioned in this guide. The environment variables in the guides have been renamed according to the output of `rhoas generate-config` command.

jackdelahunt:
Thank you, I wasn't sure if the issue was with the guide or the code so I changed the guide just in case there are other guides for the same code to keep them in sync, but this seems to make more sense :+1: .

rkpattnaik780 (approved):

wtrocki:
I have made release. We will need to include that changes on sync. 

CC @redhat-developer/app-services-docs

PR #529, opened 2022-06-13
Update topic properties

sterobin:
Luke Chen, ready for your SME review. I've applied the policy updates from [MGDSTRM-7712](https://issues.redhat.com/browse/MGDSTRM-7712) to the topic configs doc. I've also pinged you with a few specific questions. Can you have a look through to verify this PR? Thanks.

Links:
- [Jira](https://issues.redhat.com/browse/MGDSTRM-8526)
- [Doc preview](http://file.rdu.redhat.com/~sterobin/JK-8526/README.html#ref-supported-topic-properties_configuring-topics)

### DIFF
--- a/docs/kafka/topic-configuration-kafka/README.adoc
+++ b/docs/kafka/topic-configuration-kafka/README.adoc
@@ -581,10 +581,10 @@ h|Type
 |Integer
 
 h|Default value
-|`1073741824` (bytes, 1 GB)
+|`1073741824` (bytes)
 
 h|Supported values
-| 52428800 bytes (50 MB) or greater
+| 52428800 bytes or greater
 
 h|Kafka property name
 |`segment.bytes`
@@ -723,7 +723,7 @@ h|Type
 |Integer
 
 h|Default value
-|`10485760` (bytes, 10 MB)
+|`10485760` (bytes)
 
 h|Kafka property name
 |`segment.index.bytes`
--- a/docs/kafka/topic-configuration-kafka/README.adoc
+++ b/docs/kafka/topic-configuration-kafka/README.adoc
@@ -164,7 +164,7 @@ Partitions are distinct lists of messages within a topic. Partitions are the mai
 [cols="25%,75%"]
 |===
 h|Editable
-|Yes (increase only)
+|Yes
 
 h|Type
 |Integer
@@ -173,7 +173,7 @@ h|Default value
 |`1`
 
 h|Supported values
-|[`1,...`]
+|1 or greater
 
 h|Kafka property name
 |`num.partitions`
@@ -205,6 +205,8 @@ h|Type
 h|Default value
 |`3`
 
+(For trial Kafka instances, the default value is `1`.)
+
 h|Kafka property name
 |`replication.factor`
 |===
@@ -227,6 +229,8 @@ h|Type
 h|Default value
 |`2`
 
+(For trial Kafka instances, the default value is `1`.)
+
 h|Kafka property name
 |`min.insync.replicas`
 |===
@@ -247,10 +251,7 @@ h|Type
 |Long
 
 h|Default value
-|`604800000 ms` (7 days)
-
-h|Supported values
-| `milliseconds`, `seconds`, `days`, `months`, `years`, `Unlimited`
+|`604800000` (milliseconds, 7 days)
 
 h|Kafka property name
 |`retention.ms`
@@ -272,10 +273,7 @@ h|Type
 |Long
 
 h|Default value
-|`Unlimited`
-
-h|Supported values
-| `bytes`, `kilobytes`, `megabytes`, `gigabytes`, `terabytes`, `Unlimited`
+|`-1` (no retention size limit)
 
 h|Kafka property name
 |`retention.bytes`
@@ -301,7 +299,7 @@ h|Type
 |Integer
 
 h|Default value
-|`1 MB`
+|`1048588` (bytes)
 
 h|Kafka property name
 |`max.message.bytes`
@@ -345,10 +343,54 @@ h|Type
 |Long
 
 h|Default value
-|`9223372036854775807 ms`
+|`9223372036854775807` (milliseconds)
 
 h|Kafka property name
 |`message.timestamp.difference.max.ms`
+|===
+--
+
+Message format version::
++
+--
+Message format version is the `ApiVersion` value that the broker uses to append messages to topics. This value must be a valid `ApiVersion` value, such as 0.10.0, 1.1, 2.8, or 3.0.
+
+.Property attributes
+[cols="25%,75%"]
+|===
+h|Editable
+|No
+
+h|Type
+|String
+
+h|Default value
+|`3.0-IV1`
+
+h|Kafka property name
+|`message.format.version`
+|===
+--
+
+Message down-conversion::
++
+--
+Message down-conversion determines whether the broker can convert the `message.format.version` property value to an older version for consumers that require an older message format version. By default, this property is enabled in order to avoid an `UNSUPPORTED_VERSION` error for consumption requests from older clients. If this property adds excessive load to your broker, you can disable it.
+
+.Property attributes
+[cols="25%,75%"]
+|===
+h|Editable
+|Yes
+
+h|Type
+|Boolean
+
+h|Default value
+|`true`
+
+h|Kafka property name
+|`message.downconversion.enable`
 |===
 --
 
@@ -374,30 +416,6 @@ h|Kafka property name
 |`compression.type`
 |===
 --
-
-
-Message format version::
-+
---
-Message format version is the `ApiVersion` value that the broker uses to append messages to topics. This value must be a valid `ApiVersion` value, such as 0.8.2, 0.9.0.0, or 0.10.0.
-
-.Property attributes
-[cols="25%,75%"]
-|===
-h|Editable
-|No
-
-h|Type
-|String
-
-h|Default value
-|`2.7-IV2`
-
-h|Kafka property name
-|`message.format.version`
-|===
---
-
 
 === Log
 
@@ -439,13 +457,13 @@ Delete retention time is the amount of time that deletion tombstone markers are 
 [cols="25%,75%"]
 |===
 h|Editable
-|No
+|Yes
 
 h|Type
 |Long
 
 h|Default value
-|`86400000 ms` (1 day)
+|`86400000` (milliseconds, 1 day)
 
 h|Kafka property name
 |`delete.retention.ms`
@@ -483,16 +501,38 @@ Minimum compaction lag time is the minimum time a message remains uncompacted in
 [cols="25%,75%"]
 |===
 h|Editable
-|No
+|Yes
 
 h|Type
 |Long
 
 h|Default value
-|`0 seconds`
+|`0` (milliseconds)
 
 h|Kafka property name
 |`min.compaction.lag.ms`
+|===
+--
+
+Maximum compaction lag time::
++
+--
+Maximum compaction lag time is the maximum time a message remains uncompacted in a log. This property applies only when the topic cleanup policy is set to `Compact` or `Compact, Delete`.
+
+.Property attributes
+[cols="25%,75%"]
+|===
+h|Editable
+|Yes
+
+h|Type
+|Long
+
+h|Default value
+|`9223372036854775807` (milliseconds)
+
+h|Kafka property name
+|`max.compaction.lag.ms`
 |===
 --
 
@@ -515,7 +555,7 @@ h|Type
 |Boolean
 
 h|Default value
-|`Disabled`
+|`false`
 
 h|Kafka property name
 |`unclean.leader.election.enable`
@@ -535,13 +575,16 @@ Log segment size is the size of the log segment files that constitute the log. L
 [cols="25%,75%"]
 |===
 h|Editable
-|No
+|Yes
 
 h|Type
 |Integer
 
 h|Default value
-|`1 GB`
+|`1073741824` (bytes, 1 GB)
+
+h|Supported values
+| 52428800 bytes (50 MB) or greater
 
 h|Kafka property name
 |`segment.bytes`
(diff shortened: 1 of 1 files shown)


docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -375,7 +399,8 @@ h|Kafka property name
 |===
 --
 
-
+////
+// Removing for now per epic brief: "message.format.version: Don't allow setting it even to default". (Stetson, 13 June 2022)
```
sterobin, line 403: Luke, I've removed this `message.format.version` property per the epic brief, but see my last comment about message down-conversion, which relates to this and thus makes me wonder why we wouldn't keep this property in support.

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -227,6 +227,8 @@ h|Type
 h|Default value
 |`2`
 
+(For trial Kafka instances, the default value is `1`.)
+
 h|Kafka property name
 |`min.insync.replicas`
 |===
```
sterobin, line 234: Luke, not part of this PR, but have a look at the "Retention time" and "Retention size" below at lines 237 and 262. Note how they are of type `Long` but either the default value or the supported values or both show just the measurements supported as string values, or the value `Unlimited`. I didn't notice that before, but seems odd for a `Long` type to mention strings for supported or default values. 

How about we simply remove the "Supported values" row in the two tables and change the default value for Retention size from `Unlimited` to `9223372036854775807 ms` (what Kafka considers unlimited)? That work or is there something I've misunderstood?

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -535,14 +582,17 @@ Log segment size is the size of the log segment files that constitute the log. L
...
+|Yes
 
 h|Type
 |Integer
 
 h|Default value
 |`1 GB`
 
+h|Supported values
+| 50 MB or greater
```
showuon, line 594: Because the unit is `byte`, it should change to `52428800` instead.
ex: 52428800 (50 MB) or greater

WDYT?

showuon: And the default value `1 GB` is not correct. It should be `1073741824` bytes = 1GB

showuon (commented):
@sterobin , Overall LGTM. left some comments. I'll check the message format in Kafka, and get back to you later. Thanks.

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: @showuon , are you sure about `-1`? The value that Kafka seems to use for "unlimited" for this property and for many others is `9223372036854775807`, which is thousands of years and virtually unlimited. To have -1 bytes doesn't sound unlimited, in fact it sounds like it would disallow you to have anything. So I'm confused there.

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -352,6 +354,28 @@ h|Kafka property name
 |===
 --
 
+Message down-conversion::
```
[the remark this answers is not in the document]
sterobin: @showuon Sure thing. But just to clarify again, my specific confusion is that we are with this update both removing the `message.format.version` and adding `message.downconversion.enable` set to `true` by default, which therefore hinges on the message format property being up to date or not, which we're removing and no allowing users to update. So it's just kind of odd and a bit of circular logic from what I'm understanding. Thanks.

docs/kafka/topic-configuration-kafka/README.adoc
showuon: You're right. Let me re-phrase it.

- retention.bytes
defualt value -> `-1` , which means **disable** the size control for log retention.
supported value row should be removed

docs/kafka/topic-configuration-kafka/README.adoc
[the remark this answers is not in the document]
showuon: I'm not sure I got your idea. But here's my thought:

> My understanding is that as an alternative to using this down-conversion property, users can just ensure that their clients get upgraded to the correct message.format.version to avoid down-conversion

Yes, that's true, if the users upgrade their producer first to allow newer version of `message.format.version`. But that's not under our control. Besides, there might be some other cases we don't know that needs down-conversion. 
From my understanding, the disadvantage of allowing down-conversion is it will consume CPU resources. But on the other hand, if provide flexibility to users when they need to consume old message versions.

My thought is:
1. Maybe you're right, since we don't allow users to edit `message.format.version`, there will not have `down-conversion` happen. If that's true, then this down-conversion property won't hurt our server
2. If there are cases that user still needs to do down-conversion, we can find it if the CPU load is high, and then change the limit. But before that, I think providing the flexibility to users is a better option.

Does that make sense?

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: Ah, I follow now. I've updated. Thanks.

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: @showuon So you're sure it's okay to remove this property since we've also introduced down-conversion, which is related? Per our other conversation in that comment.

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -339,19 +335,41 @@ Maximum message timestamp difference is the maximum difference allowed between t
...
 
 h|Kafka property name
 |`message.timestamp.difference.max.ms`
 |===
 --
 
+Message down-conversion::
++
+--
+Message down-conversion determines whether the broker can convert the `message.format.version` property value to an older version for consumers that require an older message format version. By default, this property is enabled in order to avoid an `UNSUPPORTED_VERSION` error for consumption requests from older clients. If this property adds excessive load to your broker, you can disable it as needed.
```
jbyrne-redhat, line 354: Can remove "as needed".

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -483,19 +502,41 @@ Minimum compaction lag time is the minimum time a message remains uncompacted in
...
+[cols="25%,75%"]
+|===
+h|Editable
+|Yes
+
+h|Type
+|Long
+
+h|Default value
+|`9223372036854775807` (milliseconds)
```
jbyrne-redhat, line 533: Just checking that this value is actually correct? I used an online converter to check what it corresponds to, and it came out at, like, 290 million years? :D

jbyrne-redhat (commented):
Just a couple of comments, @sterobin.

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -227,6 +227,8 @@ h|Type
 h|Default value
 |`2`
 
+(For trial Kafka instances, the default value is `1`.)
```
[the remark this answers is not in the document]
sterobin: @jbyrne-redhat Good point. @showuon Can you verify? Is it true that `replication.factor` also defaults to `1` for trial instances, or is it just this `min.insync.replicas` property?

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: Ha! Yeah, I know. Yes, that's correct. It's the default Long value that kafka uses when it basically means "unlimited". You'll see this number come up all over the place in [kafka properties](https://kafka.apache.org/documentation/#topicconfigs).

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -725,7 +772,7 @@ h|Type
 |Long
 
 h|Default value
-|`9223372036854775807 ms`
+|`9223372036854775807` (milliseconds)
```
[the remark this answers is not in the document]
sterobin: Ditto

wtrocki:
In long run it will be good to actually generate our limits from api. If you think this is something we want I can create jira for it

docs/kafka/topic-configuration-kafka/README.adoc
jbyrne-redhat: Yeah, I see what you're saying, but I think _excessive_ already implies that this high load is unwanted.

sterobin: Sure thing. Removed.

docs/kafka/topic-configuration-kafka/README.adoc
showuon: Yes, `message.format.version` property is uneditable. So does that mean we should remove it on the doc? I'm not sure.

docs/kafka/topic-configuration-kafka/README.adoc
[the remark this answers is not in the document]
showuon: aha, good catch! Yes, `replication.factor` default to `1` for trial instance.

docs/kafka/topic-configuration-kafka/README.adoc
showuon: @sterobin , this is the only question I have, otherwise, LGTM. Thanks.

sterobin: @showuon Well, there's a difference between "uneditable" vs. what the Jira says for this property, which is "Don't allow setting it **even to default**" So not not enable users to even set the default means that we don't support this property, and therefore it should be removed. But if that wording is innaccurate, and if what we really mean is "This property is uneditable" and the default = 2.7-IV2`, then yes, we can should keep.

Is there a way you can confirm whether this property is uneditable vs. unsupported? If supported but simply uneditable, can you verify if the default value is still `2.7-IV2`?

Thanks.

docs/kafka/topic-configuration-kafka/README.adoc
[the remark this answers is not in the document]
sterobin: Fixed. Great catch @jbyrne-redhat !

docs/kafka/topic-configuration-kafka/README.adoc
showuon: It's `uneditable`, and default value is `3.0-IV1`. The default value should not be updated because this config is deprecated now. Thanks.

showuon: Sorry, on second thought, I think you were right. Since we don't allow them to set it, even to default value. So it's `unsupported`. And I agree we don't have to put it in doc. Thanks.

sterobin: @showuon ack. I've restored the property and updated that default. Thank you!

showuon: Anyway, since you've added it as `uneditable`, I don't think it is wrong. After all, they are the `read-only` property. So, let's keep it. Thanks.

showuon (approved):
LGTM! Thanks for the update!

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -301,7 +299,7 @@ h|Type
 |Integer
 
 h|Default value
-|`1 MB`
+|`1000000` (bytes, 1 MB)
```
asanzgom, line 302: In the case of Trial Kafka Instances, default value is unlimited

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -317,7 +315,7 @@ Message timestamp type determines whether the timestamp is generated when the me
 [cols="25%,75%"]
 |===
 h|Editable
-|No
+|Yes
```
asanzgom, line 318: Is this change taking effect on non-trial Kafka Instances? On trial ones it's not Editable by default...

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -339,24 +337,23 @@ Maximum message timestamp difference is the maximum difference allowed between t
...
```
asanzgom, line 340: Is this change taking effect on non-trial Kafka Instances? On trial ones it's not Editable by default...

sterobin: @showuon Please confirm.

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: @showuon Please confirm.

docs/kafka/topic-configuration-kafka/README.adoc
mmusil: Not sure where @asanzgom got unlimited, but this is the default value for max message bytes that I can see
image

mmusil: Perhaps he did not expand those collapsed lines and thought it was meant for Retention size.
image

docs/kafka/topic-configuration-kafka/README.adoc
mmusil: This is not editable for standard Kafka instances either. Tested on stage.

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: @mmusil Ack.

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: @mmusil Ack, changed back to No.

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: @mmusil Ack, changing back to No.

sterobin:
@mmusil Thanks so much for the comments! I addressed the things you pointed out. With that taken care of, do you approve?

mmusil (approved):
Looks good, thanks for the changes.

docs/kafka/topic-configuration-kafka/README.adoc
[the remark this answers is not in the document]
sterobin: @showuon Fixed, and I removed 1 MB since that's technically no longer true given the new value. So now it's `1048588 (bytes)`. Thanks.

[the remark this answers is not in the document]
showuon: You're correct! Thanks for the update. I have no more comments.

jbyrne-redhat (approved):

PR #535, opened 2022-06-21
Update style issues in topic config doc

sterobin:
@jbyrne-redhat Ready for your review. Here's the [doc preview](http://file.rdu.redhat.com/~sterobin/JK-8526/README.html). I've left a few specific comments just FYI. Thanks.

### DIFF
--- a/docs/kafka/topic-configuration-kafka/README.adoc
+++ b/docs/kafka/topic-configuration-kafka/README.adoc
@@ -81,9 +81,9 @@ ifdef::context[:parent-context: {context}]
 // Purpose statement for the assembly
 [role="_abstract"]
 --
-As a developer of applications and services, you can refer to the properties of your topics in {product-kafka} to better understand how the Kafka implementation is managed for your services. You can edit certain topic properties according to the needs and goals of your services. Kafka topics contain the data (events) that applications produce or consume, so the way the topics are configured affects how data is stored and exchanged between applications.
+As a developer of applications and services, you can refer to the properties of your topics in {product-long-kafka} to better understand how the Kafka implementation is managed for your services. You can edit certain topic properties according to the needs and goals of your services. Kafka topics contain the data (events) that applications produce or consume, so the way the topics are configured affects how data is stored and exchanged between applications.
 
-In addition, as a developer, you can use the {product-kafka} web console to check Kafka topics in {product-kafka} for matching schemas in {product-long-registry}. When you use a schema with your Kafka topic, the schema ensures that producers publish data that conforms to a certain structure and compatibility policy. The schema also helps consumers parse and interpret the data from a topic as it is meant to be read. Using the web console to quickly identify matches between Kafka topics and schemas means that you or others in your organization don't need to inspect client application code to check for schema usage. If you don’t have any {registry} instances, or don’t have a schema that matches your topic, the console provides links to {product-registry}, where you can create instances and schemas.
+In addition, as a developer, you can use the {product-kafka} web console to check Kafka topics in {product-kafka} for matching schemas in {product-long-registry}. When you use a schema with your Kafka topic, the schema ensures that producers publish data that conforms to a certain structure and compatibility policy. The schema also helps consumers parse and interpret the data from a topic as it is meant to be read. Using the web console to quickly identify matches between Kafka topics and schemas means that you or others in your organization do not need to inspect client application code to check for schema usage. If you don’t have any {registry} instances, or don’t have a schema that matches your topic, the console provides links to {product-registry}, where you can create instances and schemas.
 --
 
 //Additional line break to resolve mod docs generation error, not sure why. Leaving for now. (Stetson, 20 May 2021)
@@ -92,7 +92,7 @@ In addition, as a developer, you can use the {product-kafka} web console to chec
 == Reviewing and editing topic properties in {product-kafka}
 
 [role="_abstract"]
-Use the {product-kafka} web console to select a topic in your Kafka instance and review the topic properties. You can adjust the editable topic properties as needed.
+Use the {product-long-kafka} web console to select a topic in your Kafka instance and review the topic properties. You can adjust the editable topic properties as needed.
 
 As an alternative to using the {product-kafka} web console, you can use the `rhoas` command-line interface (CLI) to update certain topic properties, as shown in the following example command:
 
@@ -102,15 +102,13 @@ As an alternative to using the {product-kafka} web console, you can use the `rho
 rhoas kafka topic update --name my-kafka-topic --retention-ms 704800000
 ----
 
-For a list of topic properties that you can update using the CLI, see the `rhoas kafka topic update` entry in the {base-url-cli}{command-ref-url-cli}[_CLI command reference (rhoas)_^].
+For a list of topic properties that you can update using the CLI, see the `rhoas kafka topic update` entry in the {base-url-cli}{command-ref-url-cli}[CLI command reference (rhoas)^].
 
 .Prerequisites
-* You are logged in to the {product-kafka} web console.
-* You have created a Kafka instance with at least one Kafka topic in {product-kafka}.
+* You have created a Kafka instance with at least one Kafka topic in {product-kafka}. To learn how to do this, see {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^].
 
 .Procedure
-. In the {product-kafka} web console, go to *Streams for Apache Kafka* > *Kafka Instances*.
-. Click the Kafka instance that contains the topic you want to configure.
+. In the {product-kafka} {service-url-kafka}[web console], click *Kafka Instances* and select the Kafka instance that contains the topic you want to configure.
 . Select the *Topics* tab.
 . Select the options icon (three vertical dots) for the relevant topic and click *Edit* to review the current topic properties, and adjust any editable topic properties as needed.
 . Click *Save* to finish.
@@ -119,20 +117,20 @@ NOTE: You can also edit topic properties by selecting the *Properties* tab withi
 
 [role="_additional-resources"]
 .Additional resources
-* {base-url}{getting-started-url-kafka}[_Getting started with {product-long-kafka}_^]
-* {base-url}{getting-started-rhoas-cli-url-kafka}[_Getting started with the rhoas CLI for OpenShift Streams for Apache Kafka_^]
-* {base-url-cli}{command-ref-url-cli}[_CLI command reference (rhoas)_^]
+* {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^]
+* {base-url}{getting-started-rhoas-cli-url-kafka}[Getting started with the rhoas CLI for OpenShift Streams for Apache Kafka^]
+* {base-url-cli}{command-ref-url-cli}[CLI command reference (rhoas)^]
 
 
 [id="ref-supported-topic-properties_{context}"]
 == Supported topic properties in {product-kafka}
 
 [role="_abstract"]
-The following Kafka topic properties are supported in {product-kafka}. Each listed topic property indicates whether the property is editable or read only, and includes other relevant property attributes for your reference.
+The following Kafka topic properties are supported in {product-long-kafka}. Each listed topic property indicates whether the property is editable or read only, and includes other relevant property attributes for your reference.
 
 === Core configuration
 
-These properties determine the identity and core behavior of the topic. Before deploying your topic, enter all core configuration properties.
+The following topic properties determine the identity and core behavior of the topic. Before deploying your topic, enter all core configuration properties.
 
 Name::
 +
@@ -283,7 +281,7 @@ h|Kafka property name
 
 === Messages
 
-These properties control how your messages are handled in the Kafka instance.
+The following topic properties control how your messages are handled in the Kafka instance.
 
 Maximum message bytes::
 +
@@ -420,7 +418,7 @@ h|Kafka property name
 
 === Log
 
-These properties define how your log is handled.
+The following topic properties define how your log is handled.
 
 NOTE: Messages are continually appended to the partition log and are assigned their offset.
 
@@ -539,7 +537,7 @@ h|Kafka property name
 
 === Replication
 
-These properties control the behavior of your replicas. Each of these properties impacts every replica created in the topic.
+The following topic properties control the behavior of your replicas. Each of these properties impacts every replica created in the topic.
 
 Unclean leader election::
 +
@@ -565,7 +563,7 @@ h|Kafka property name
 
 === Cleanup
 
-These properties control the cleanup processing of the log.
+The following topic properties control the cleanup processing of the log.
 
 Log segment size::
 +
@@ -685,7 +683,7 @@ h|Kafka property name
 
 === Index
 
-These properties control the indexing of the log.
+The following topic properties control the indexing of the log.
 
 Index interval size::
 +
@@ -733,7 +731,7 @@ h|Kafka property name
 
 === Flush
 
-These properties control the frequency of the flushing of the log.
+The following topic properties control the frequency of the flushing of the log.
 
 Flush interval messages::
 +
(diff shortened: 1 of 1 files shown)


docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -101,15 +101,13 @@ As an alternative to using the {product-kafka} web console, you can use the `rho
 rhoas kafka topic update --name my-kafka-topic --retention-ms 704800000
 ----
 
-For a list of topic properties that you can update using the CLI, see the `rhoas kafka topic update` entry in the {base-url-cli}{command-ref-url-cli}[_CLI command reference (rhoas)_^].
+For a list of topic properties that you can update using the CLI, see the `rhoas kafka topic update` entry in the {base-url-cli}{command-ref-url-cli}[CLI command reference (rhoas)^].
 
 .Prerequisites
-* You are logged in to the {product-kafka} web console.
-* You have created a Kafka instance with at least one Kafka topic in {product-kafka}.
+* You have created a Kafka instance with at least one Kafka topic in {product-kafka}. To learn how to do this, see {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^].
```
sterobin, line 107: @jbyrne-redhat Only reason I added that cross ref was that Steve did so in the schema sections at the end, which merits them given the different context. For consistency, I've added the same here. Also for consistency, I used his same wording "To learn more" instead of what I normally use ("For more information ..." ) I think Steve was trying to avoid the repetition of "doing x, y, z.  For more information about doing x, y, z, see ...."  I typically try to reword, but for now just stuck with his alternative, which works.

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -118,20 +116,20 @@ NOTE: You can also edit topic properties by selecting the *Properties* tab withi
...
-* {base-url}{getting-started-url-kafka}[_Getting started with {product-long-kafka}_^]
-* {base-url}{getting-started-rhoas-cli-url-kafka}[_Getting started with the rhoas CLI for OpenShift Streams for Apache Kafka_^]
-* {base-url-cli}{command-ref-url-cli}[_CLI command reference (rhoas)_^]
+* {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^]
+* {base-url}{getting-started-rhoas-cli-url-kafka}[Getting started with the rhoas CLI for OpenShift Streams for Apache Kafka^]
+* {base-url-cli}{command-ref-url-cli}[CLI command reference (rhoas)^]
 
 
 [id="ref-supported-topic-properties_{context}"]
 == Supported topic properties in {product-kafka}
```
sterobin, line 125: @jbyrne-redhat All topic properties have a table of "Property attributes" with that same heading. As opposed to "Property attributes for ___ property", which I think is excessive. And none of the tables have intro statements, again excessive.

Both of these facts are intentional given the nature of the content as reference content, like API docs etc, where we similarly have many tables or blocks with the same "Required parameters" heading or "Example request" / "Example response" etc., not specific to every single one.

Just noting in passing in light of the style focus lately. Open to any other thoughts.

jbyrne-redhat: Yep, I agree with all of that. You're right that expanded captions and the addition of intro sentences would be excessive.  And I think you're right to treat reference content a little differently.

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -282,7 +280,7 @@ h|Kafka property name
 
 === Messages
 
-These properties control how your messages are handled in the Kafka instance.
+The following topic properties control how your messages are handled in the Kafka instance.
```
jbyrne-redhat, line 283: I wonder if we can eliminate passive voice from this and a couple of the subsequent intro sentences? Per 2.4 [here](https://docs.google.com/document/d/1TERH4XRH6La44hUQH88LUaax-35JH3dLXmceGJoLv7I/edit#) we cautioned against not using active voice if the agent of an action is unknown or unimportant, when it makes a sentence awkward or clunky, or when it introduces anthropomorphism. But I'm not sure that would be the case with a couple of these.

For example, it seems reasonable to say "The following topic properties control how the Kafka instance handles messages" and "The following topic properties define how the Kafka instance handles your log."

WDYT?

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -783,55 +781,54 @@ h|Kafka property name
 * https://kafka.apache.org/documentation/#topicconfigs[Topic-Level Configs^] in Kafka documentation
 
 [id="con-using-kafka-topics-with-registry-schemas_{context}"]
-== Using topics in {product-long-kafka} with schemas in {product-long-registry}
+== Schemas in {product-registry} for topics in {product-kafka}
```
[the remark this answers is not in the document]
jbyrne-redhat: Yeah, per my previous comment, this section was added by me. The concept module is effectively acting as an assembly within the main assembly, hence the gerund title. I might even have discussed this approach with you prior to doing it. It was preferable to creating a standalone guide, because the content fit quite well within this one.

So, I think we'd need to restore the original title if we keep this content where it is.

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -783,55 +781,54 @@ h|Kafka property name
...
+By default, a Kafka topic that you create in {product-long-kafka} can store any kind of data. The topic does not validate the structures of messages that it stores. However, as a developer of applications and services, you might want to define the structure of the data for messages stored in a given topic, and ensure that producers and consumers use this structure. To achieve this goal, you can use schemas that you upload to registry instances in {product-long-registry} with your Kafka topics. {product-registry} is a cloud service that enables you to manage schema and API definitions in your applications without having to install, configure, run, and maintain your own registry instances.
```
jbyrne-redhat, line 787: Thanks for fixing the things from here down - these were all me :)

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -783,55 +781,54 @@ h|Kafka property name
...
-* You've created a Kafka instance with at least one topic in {product-kafka}. To learn how to do this, see {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^].
+* You have created a Kafka instance with at least one topic in {product-kafka}. To learn how to do this, see {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^].
 * You understand how to create a {registry} instance and upload a schema to be used by client applications. To learn how to do this, see {base-url}{getting-started-url-registry}[Getting started with {product-long-registry}^].
 * You have access to at least one instance in {registry} that you can check for schemas that match your topic.
 
 .Procedure
-. In the {product-kafka} web console, go to *Streams for Apache Kafka* > *Kafka Instances*. Click the name of the Kafka instance that contains the topic that you want to check for matching schemas in {product-registry}.
+. In the {product-kafka} {service-url-kafka}[web console], click *Kafka Instances* and select the name of the Kafka instance that contains the topic that you want to check for matching schemas in {product-registry}.
```
jbyrne-redhat, line 811: Need to include a caret here: [web console ^].

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -783,55 +781,54 @@ h|Kafka property name
...
 
 .Prerequisites
-* You're logged in to the {product-kafka} web console.
-* You've created a Kafka instance with at least one topic in {product-kafka}. To learn how to do this, see {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^].
+* You have created a Kafka instance with at least one topic in {product-kafka}. To learn how to do this, see {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^].
 * You understand how to create a {registry} instance and upload a schema to be used by client applications. To learn how to do this, see {base-url}{getting-started-url-registry}[Getting started with {product-long-registry}^].
 * You have access to at least one instance in {registry} that you can check for schemas that match your topic.
 
 .Procedure
-. In the {product-kafka} web console, go to *Streams for Apache Kafka* > *Kafka Instances*. Click the name of the Kafka instance that contains the topic that you want to check for matching schemas in {product-registry}.
```
jbyrne-redhat, line 814: Surely I didn't write "go to"!? :roll_eyes: Thanks for fixing :relaxed:

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -783,55 +781,54 @@ h|Kafka property name
...
+. In the {product-kafka} {service-url-kafka}[web console], click *Kafka Instances* and select the name of the Kafka instance that contains the topic that you want to check for matching schemas in {product-registry}.
 . On the *Topics* page, click the name of the topic that you want to check.
 . Click the *Schemas* tab.
-. In the *{registry} instance* drop-down menu, select a {registry} instance to check for schemas that have IDs that match the name of the topic.
+. In the *{registry} instance* list, select a {registry} instance to check for schemas that have IDs that match the name of the topic.
 +
 The *Schemas* tab shows any schemas registered to the selected {registry} instance that match the topic.
 +
-NOTE: Although the drop-down menu shows all {registry} instances in your organization, you can see schema information *only* for instances that you own or have been granted access to.
+NOTE: Although the instance list shows all {registry} instances in your organization, you can see schema information only for instances that you own or have been granted access to.
```
jbyrne-redhat, line 818: I'm glad we're adopting this - much cleaner.

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -783,55 +781,54 @@ h|Kafka property name
...
 * You understand how to create a {registry} instance and upload a schema to be used by client applications. To learn how to do this, see {base-url}{getting-started-url-registry}[Getting started with {product-long-registry}^].
 * You have access to at least one instance in {registry} that you can check for schemas that match your topic.
 
 .Procedure
-. In the {product-kafka} web console, go to *Streams for Apache Kafka* > *Kafka Instances*. Click the name of the Kafka instance that contains the topic that you want to check for matching schemas in {product-registry}.
+. In the {product-kafka} {service-url-kafka}[web console], click *Kafka Instances* and select the name of the Kafka instance that contains the topic that you want to check for matching schemas in {product-registry}.
 . On the *Topics* page, click the name of the topic that you want to check.
 . Click the *Schemas* tab.
-. In the *{registry} instance* drop-down menu, select a {registry} instance to check for schemas that have IDs that match the name of the topic.
+. In the *{registry} instance* list, select a {registry} instance to check for schemas that have IDs that match the name of the topic.
```
jbyrne-redhat, line 814: I'm glad we're adopting this - much cleaner.

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -783,55 +781,54 @@ h|Kafka property name
...
 +
 The *Schemas* tab shows any schemas registered to the selected {registry} instance that match the topic.
 +
-NOTE: Although the drop-down menu shows all {registry} instances in your organization, you can see schema information *only* for instances that you own or have been granted access to.
+NOTE: Although the instance list shows all {registry} instances in your organization, you can see schema information only for instances that you own or have been granted access to.
 
-. If the *Schemas* tab shows the schema types that you want associated with your topic, you're done. You don't need to complete the remainder of this procedure.
+. If the *Schemas* tab shows the schema types that you want associated with your topic, you do not need to complete the remainder of this procedure.
```
jbyrne-redhat, line 820: :+1:

docs/kafka/topic-configuration-kafka/README.adoc
```diff
@@ -853,29 +850,28 @@ The *Schemas* tab now shows the name of the matching schema that you uploaded.
...
 
 .Prerequisites
-* You're logged in to the {product-kafka} web console.
-* You've created a Kafka instance with at least one Kafka topic in {product-kafka}. To learn how to do this, see {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^].
+* You have created a Kafka instance with at least one Kafka topic in {product-kafka}. To learn how to do this, see {base-url}{getting-started-url-kafka}[Getting started with {product-long-kafka}^].
 * You understand how to create a {registry} instance and upload a schema to be used by client applications. To learn how to do this, see {base-url}{getting-started-url-registry}[Getting started with {product-long-registry}^].
 
 .Procedure
-. In the {product-kafka} web console, go to *Streams for Apache Kafka* > *Kafka Instances*. Click the name of the Kafka instance that contains the topic that you want to check for matching schemas in {product-registry}.
+. In the {product-kafka} {service-url-kafka}[web console], click *Kafka Instances* and select the name of the Kafka instance that contains the topic that you want to check for matching schemas in {product-registry}.
```
jbyrne-redhat, line 862: Same comment as https://github.com/redhat-developer/app-services-guides/pull/535/files#r904878992

jbyrne-redhat (commented):
@sterobin This is a really sound set of updates and completely in line with our recent style discussions - thank you for doing this. Just a few comments and replies for you to consider.

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: Well that's handy. Ack.

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: Good point. I reviewed all intro sentences and revised the ones with passive voice to be active. I also reviewed briefly the property descriptions as well, and many of them use passive too, but imo, revising those ones is a bit more debatable and subjective and might fall into the category of more harm than good. But the intro sentences for sure are an easy fix, so done.

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: Shoot, forgot! I've fixed all 3 instances. Thank you!

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: Agreed!

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: Oh, that's not you, it's actually more me, and it was intentional. I bet you copied the standard we/I established early on. We say "go to" everywhere in docs today where we have a path for users to follow, including in my docs. As opposed to "navigate to" of just "select X > Y > Z". IBM says to just say "select", "click", "expand", whatever, per [Menus and navigation](https://www.ibm.com/docs/en/ibm-style?topic=elements-menus-navigation) section. 

However, the deviation to go with "go to" was intentional, because we found on my previous team (BA) that IBM's ruling of "select", "click", "expand" doesn't always work with navigations because the first thing in the series could be something you click, the second something you select, and the third something you expand. So picking one verb for the navigation series just wasn't scalable or sensical many times. Whereas "navigate to" or "go to" works every time. So we adopted that going forward years ago, and I guess I brought that with me :)

That said, this standard was admittedly brought into play mostly by me, and I think Ben followed, then everyone else likewise followed, etc. So it sounds like we should revisit it and my logic above for it as a team and decide whether to keep doing it or ditch it and add it to that style table checklist of things to rectify.

jbyrne-redhat: Ah, I'm sorry for picking on this phrase specifically - I was sure I was the one to have written it. I actually forgot how much it crops up elsewhere. 

When I read it yesterday, it suddenly jumped out to me as very vague in the action it's communicating to users. But the logic you described in your comment makes total sense. I'd even say that it suits the informal voice of our quick starts, where "click", "select", etc. get quite stuffy.

So we'll maybe take a look at this again if we have more style discussions, but I definitely don't see it as a burning issue. Thanks for taking the time to explain the origin of this.

jbyrne-redhat (approved):

docs/kafka/topic-configuration-kafka/README.adoc
sterobin: Ack, thanks.

PR #536, opened 2022-06-22
docs: Add new attribute for service contexts guide

jbyrne-redhat:


jbyrne-redhat:
> Looks good! Also, just a reminder that you'll need to make the applicable change to each attributes file in the downstream repos. We still don't have a process set up yet to synchronize the downstream attributes across repos :-(.

Thanks, @bhardesty. Yes, thanks for reminder - I'll be adding the equivalent attribute to the each of the downstream repos.

PR #537, opened 2022-06-23
docs: Update Quarkus and Node.js quick starts to better describe manual connection use case

jbyrne-redhat:
**Notes for reviewers**

- An HTML preview is available for review [here](http://file.emea.redhat.com/~jbyrne/JK-8867/README.html)
- This guide requires some updates to meet our style guidelines. I plan to make these in a separate PR.


wtrocki (approved):

docs/kafka/nodejs-kafka/README.adoc
```diff
@@ -74,14 +74,18 @@ END GENERATED ATTRIBUTES
 ////
 
 [id="chap-using-nodejs"]
-= Using Node.js applications with Kafka instances in {product-long-kafka}
+= Manually connecting Node.js applications to Kafka instances in {product-long-kafka}
```
DuncanDoyle, line 77: I wonder if we should use the word  "Manually" (as it might have a "negative" ring to it). We can say that that is implicit and we want to call out when we connect "automatically.

docs/kafka/nodejs-kafka/README.adoc
```diff
@@ -91,22 +95,20 @@ endif::[]
...
 
-[NOTE]
-====
-The example Node.js application in this quick start uses the https://kafka.js.org/[KafkaJS^] client by default. If you want to use the https://github.com/blizzard/node-rdkafka[node-rdkafka^] client, you must install some development tools locally on your computer, or use Docker to run a specified container image and configure a development environment. To learn more, see the https://github.com/nodeshift-starters/reactive-example/tree/node-rdkafka#node-rdkafka-and-kafkajs[documentation] for the example Node.js application.
-====
+NOTE: The example Node.js application in this quick start uses the https://kafka.js.org/[KafkaJS^] client by default. If you want to use the https://github.com/blizzard/node-rdkafka[node-rdkafka^] client, you must install some development tools locally on your computer, or use Docker to run a specified container image and configure a development environment. To learn more, see the https://github.com/nodeshift-starters/reactive-example/tree/node-rdkafka#node-rdkafka-and-kafkajs[documentation] for the example Node.js application.
+
```
DuncanDoyle, line 99: Instead of Docker, we could also say "container runtime, like Podman or Docker" (as Podman is our own container engine and we might want to call that out as an alternative).

docs/kafka/nodejs-kafka/README.adoc
jbyrne-redhat: @DuncanDoyle Yeah, I did hum and haw about this. The problem (as I see it) is that it's too much of a stretch to say that service contexts make things _automatic_ - just more automated. I've indicated that here in the note on line 87.

Without being able to call one approach manual or the other automatic, it's hard to draw the appropriate distinction to help users pick the right guide from our landing page. That's what these updates are geared towards. I guess I didn't see it as _such_ a bad thing that we use the quick starts to paint contexts as a favorable alternative to manual configuration work. Again, the note on line 87 is aimed at doing that.

WDYT?

docs/kafka/nodejs-kafka/README.adoc
jbyrne-redhat: Sure, yep - I'll make that update.

docs/kafka/nodejs-kafka/README.adoc
DuncanDoyle: Yeah, that explanation makes sense.

DuncanDoyle (approved):

sterobin (approved):
No comments, looks good and makes sense. I did likewise wonder about the terminology with "manually" doing this like Duncan mentioned, but as I think through it as you've explained, I suppose that does seem most logical. So good with me.

PR #539, opened 2022-06-23
fix: update examples and default values for sso token

wtrocki:
This PR changes visual elements in the docs that reference to service that we are deprecating. 

Change should be merged and published around migration, but it doesn't actually affects any examples and docs in any meaningful way.


jbyrne-redhat (approved):
These changes look good, @wtrocki. Just cautioning us against prematurely merging these updates.

wtrocki:
I think that actually we might merge that faster as those changes have pure informational function and are not subject to breaking anything now. 
We could remove those but I kept them to avoid changing approach we use in the docs right now

wtrocki:
Like if we change them now and they will be in production 1-2 weeks faster than actual migration nothing will happen.

PR #551, opened 2022-07-12
docs: Remove broken xref

jbyrne-redhat:
The xref in question was pointing to the wrong section. However, the link is not actually needed, so I simply removed it.


AedinC (approved):
Approved.

wtrocki:
LGTM

PR #553, opened 2022-07-22
Kafka instance settings doc

bhardesty:
New doc for Kafka instance settings. There's only one setting at the moment (connect re-authentication), but I tried to do this in a way that would make it simple to document any Kafka instance settings that are added in the future.

@HemaHG even though the UI for the new Settings tab isn't complete yet, I think the design was sufficient for creating the doc. However, if you think it makes more sense to review the doc after the dev work is complete, I can mark this as Draft for now.


docs/kafka/kafka-instance-settings/README.adoc
```diff
@@ -0,0 +1,152 @@
...
+* {base-url-cli}{command-ref-url-cli}[CLI command reference (rhoas)^]
+
+
+[id="ref-kafka-instance-settings_{context}"]
+== Kafka instance settings in {product-kafka}
+
+[role="_abstract"]
+You can edit the following Kafka instance settings in {product-long-kafka}.
+
+Connection re-authentication::
```
bhardesty, line 126: @DuncanDoyle Can you verify that the connection reauth definition is technically accurate? I especially want to make sure that I've stated the security risk correctly.

HemaHG:
yes, I think it will be good if we wait until the UI is ready

docs/kafka/kafka-instance-settings/README.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+As an alternative to using the {product-kafka} web console, you can use the `rhoas` command-line interface (CLI) to update certain Kafka instance settings, as shown in the following example command:
+
+.Example CLI command to disable connection reauthentication
+[source]
+----
+rhoas kafka update --reauthentication false
+----
+
+For a list of Kafka instance settings that you can update using the CLI, see the `rhoas kafka update` entry in the {base-url-cli}{command-ref-url-cli}[CLI command reference (rhoas)^].
```
jbyrne-redhat, line 103: I wonder if these lines (95-103) might fit more logically as a note, at the end of the procedure for using the web console?

docs/kafka/kafka-instance-settings/README.adoc
```diff
@@ -0,0 +1,152 @@
...
+--
+As a developer of applications and services, you can review and modify settings for your Kafka instances. By modifying these settings, you can configure your Kafka instances to suit your particular environment.
+--
+
+
+[id="proc-editing-kafka-instance-settings_{context}"]
+== Reviewing and editing Kafka instance settings in {product-kafka}
+
+[role="_abstract"]
+Use the {product-long-kafka} web console to review and adjust the Kafka instance settings as needed.
```
jbyrne-redhat, line 93: Suggest:

```
Use the {product-long-kafka} web console to review and adjust settings for a Kafka instance.
```

docs/kafka/kafka-instance-settings/README.adoc
```diff
@@ -0,0 +1,152 @@
...
+[role="_abstract"]
+You can edit the following Kafka instance settings in {product-long-kafka}.
+
+Connection re-authentication::
++
+--
+When a client connects to a Kafka instance, the session lasts for five minutes. 
+At that point, the client must reauthenticate to stay connected.
+Many Kafka clients automatically reauthenticate to remain connected,
+but some Kafka clients don't.
```
jbyrne-redhat, line 132: Think we want to avoid contractions.

docs/kafka/kafka-instance-settings/README.adoc
```diff
@@ -0,0 +1,152 @@
...
+disable the `Connection re-authentication` setting.
+
+Before disabling connection re-authentication,
+you should be aware of the security risks.
+If you disable connection re-authentication,
+and then a bad actor obtains credentials to your Kafka instance,
+they will be able to stay connected indefinitely.
+Deactivating the user account or service account will not close the connections that the bad actor has opened.
+In this scenario, you would need to add ACLs to prevent the unauthorized connections from performing any operations
+(see {base-url}{access-mgmt-url-kafka}[Managing account access in OpenShift Streams for Apache Kafka^]).
```
jbyrne-redhat, line 146: A few comments on this definition:

- Avoid use of contractions
- Spell out _Access Control List_ in full when first introducing 'ACL'. Also, I'm not sure it needs to be _ACLs_ (plural) in this case? That is, I think you could define a single ACL to prevent bad actors from performing operations?
- Write this in the present tense. Avoid _will_, _would_, and _could_.
- We all understand the term "bad actor" but I still wonder if it borders on engineering jargon? Maybe we should consider a more generic phrase such as "malicious entity" or "malicious source"? At the very least, we probably need to define _bad actor_ (that is, an entity that poses a security risk) on first use.

jbyrne-redhat:
@bhardesty I know you're waiting on the UI to be finished, but I went ahead and reviewed what you have so far. Just a few comments for you to consider.

docs/kafka/kafka-instance-settings/README.adoc
jgiardino: Just commenting here to follow this discussion regarding "bad actor". 😄 We want to use a similar message in the prompt in the UI, e.g.:
> There are potential risks associated with turning off re-authentication, where bad actors could stay connected indefinitely.

We can update the text in the UI to match whatever the final decision is here.

bhardesty: Thanks for the comments, @jbyrne-redhat. I made some edits based on them. Also, I changed "bad actor" to "attacker".

HemaHG (approved):
LGTM

PR #561, opened 2022-08-02
docs: update environment variables according to CLI outputs

rkpattnaik780:
This PR updates the environment variables according to the latest changes for `rhoas generate-config` command.

To be merged after the CLI is released.


code-examples/quarkus-kafka-quickstart/src/main/resources/application.properties
```diff
@@ -26,9 +26,9 @@ quarkus.container-image.push=false
 
 %dev.kafka.sasl.mechanism=OAUTHBEARER
 %dev.kafka.sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
-  oauth.client.id="${RHOAS_CLIENT_ID}" \
-  oauth.client.secret="${RHOAS_CLIENT_SECRET}" \
-  oauth.token.endpoint.uri="${RHOAS_OAUTH_TOKEN_URL}" ;
+  oauth.client.id="${RHOAS_SERVICE_ACCOUNT_CLIENT_ID}" \
```
wtrocki, line 29: So values we have been providing in context were totally different than values in the service account file?

wtrocki: I like new variable names anyway :D

rkpattnaik780: Yes those were different.

docs/rhoas/rhoas-service-contexts/README.adoc
```diff
@@ -275,11 +275,25 @@ $ rhoas context use --name development-context
...
 ----
-$ rhoas generate-config --type env --output-file ./producer/.env
+$ rhoas generate-config --type env --output-file ./producer/rhoas.env
+----
+
+. Append the contents of connection configuration file to the service account environment variables file.
++
+[source,shell]
+----
+$ cat ./producer/rhoas.env >> ./producer/.env
```
rkpattnaik780, line 296: Does this appending approach look good? I am not sure if quarkus can load `rhoas.env`, hence appended URLs to service account file.
@wtrocki

code-examples/quarkus-service-registry-quickstart/consumer/.env
```diff
@@ -1,6 +0,0 @@
-## Generated by rhoas generate-config
```
rkpattnaik780, line 1: I have removed this file. Want to make sure if the `.env` files were added accidentally or do we intend to keep them.

docs/rhoas/rhoas-service-contexts/README.adoc
```diff
@@ -275,11 +275,25 @@ $ rhoas context use --name development-context
 $ cd ~/app-services-guides/code-examples/quarkus-service-registry-quickstart/
 ----
 
+. Generate an environment variables file that contains the credentials for service account.
```
bhardesty, line 278:
```suggestion
. Generate an environment variables file that contains the credentials for the service account.
```

docs/rhoas/rhoas-service-contexts/README.adoc
```diff
@@ -275,11 +275,25 @@ $ rhoas context use --name development-context
...
+
 . Generate an environment variables file that contains the connection configuration information required by the producer component.
 +
 [source,shell]
 ----
-$ rhoas generate-config --type env --output-file ./producer/.env
+$ rhoas generate-config --type env --output-file ./producer/rhoas.env
+----
+
+. Append the contents of connection configuration file to the service account environment variables file.
```
bhardesty, line 292:
```suggestion
. Append the contents of the connection configuration file to the service account environment variables file.
```

docs/rhoas/rhoas-service-contexts/README.adoc
```diff
@@ -275,7 +275,7 @@ $ rhoas context use --name development-context
 $ cd ~/app-services-guides/code-examples/quarkus-service-registry-quickstart/
 ----
 
-. Generate an environment variables file that contains the credentials for service account.
+. Create a service account and save the credentials in an environment variables file.
```
jbyrne-redhat, line 278: Suggest:

"Create a service account for the Quarkus application to use. Save the credentials in an environment variables file in the directory for the producer component."

docs/rhoas/rhoas-service-contexts/README.adoc
```diff
@@ -313,15 +313,15 @@ RHOAS_SERVICE_ACCOUNT_CLIENT_ID=_<client-id>_
 RHOAS_SERVICE_ACCOUNT_CLIENT_SECRET=_<client-secret>_
 RHOAS_SERVICE_ACCOUNT_OAUTH_TOKEN_URL={sso-token-url}
 ## Generated by rhoas cli
```
jbyrne-redhat, line 315: @rkpattnaik780 Is this a duplicate line?

docs/rhoas/rhoas-service-contexts/README.adoc
```diff
@@ -275,11 +275,25 @@ $ rhoas context use --name development-context
...
+----
+
+. Append the contents of the connection configuration file to the service account environment variables file.
++
+[source,shell]
+----
+$ cat ./producer/rhoas.env >> ./producer/.env
 ----
 
 . Copy the `.env` file to the directory for the consumer component, as shown in the following Linux example:
```
jbyrne-redhat, line 299: Suggest:

```
Copy the updated `.env` file to the directory for the consumer component, as shown in the following Linux example:
```

docs/rhoas/rhoas-service-contexts/README.adoc
rkpattnaik780: Hi @jbyrne-redhat , this line is present in both the env files, and since we use `cat` to do the append, entire content is copied.

jbyrne-redhat: Ah, makes sense. It was only a small thing, but thanks for verifying.

docs/rhoas/rhoas-service-contexts/README.adoc
```diff
@@ -275,14 +275,28 @@ $ rhoas context use --name development-context
 $ cd ~/app-services-guides/code-examples/quarkus-service-registry-quickstart/
 ----
 
+. Create a service account for the Quarkus application to authenticate with the Kafka and {registry} instances in the context. Save the credentials in an environment variables file in the directory for the producer component.
++
+[source,shell]
+----
+$ rhoas service-account create --type env --output-file ./producer/.env
+----
+
```
jbyrne-redhat, line 284: @rkpattnaik780 Is there a workflow to specify an existing service account, if you don't create a new one? I think what we have here is OK for v1 of these updates, but if users can also choose an existing service account, then I'll create a follow-up issue to test and document those additional couple of steps.

jbyrne-redhat (approved):
Changes look good to me, @rkpattnaik780. I did have one question for you, but I don't see it as a blocker for this PR.

docs/rhoas/rhoas-service-contexts/README.adoc
rkpattnaik780: If users want to choose an existing service account, they will need to have the credentials in a `.env` file and use it instead.
It might get tricky, if the user has the credentials saved in a json, the format will need to be manually changed. I think the workflow can be different according to what users may prefer, so I am not sure if we should document it.

jbyrne-redhat: That makes sense, @rkpattnaik780. Maybe we'll continue to think about what we might document in that regard - even if it's just a note. In the meantime,  I don't think it should block us from merging this PR.

PR #562, opened 2022-08-02
Docs: Add attributes for HCC and Service Accounts

jbyrne-redhat:


smccarthy-ie:
Thanks @jbyrne-redhat Those URLs look good :) 

I had a quick look downstream and noticed that @bredamc added the following attributes in the RHOAD repo. Wondering do we want to add these upstream also?

`:hcc-long: {org-name} Hybrid Cloud Console`
`:hcc: Hybrid Cloud Console`

jbyrne-redhat:
> Thanks @jbyrne-redhat Those URLs look good :)
> 
> I had a quick look downstream and noticed that @bredamc added the following attributes in the RHOAD repo. Wondering do we want to add these upstream also?
> 
> `:hcc-long: {org-name} Hybrid Cloud Console` `:hcc: Hybrid Cloud Console`

Thanks, @smccarthy-ie. That makes sense. However, I'm not sure I'd have used an acronym such as `hcc` in an attribute name. I think I'd favour a more user-friendly (and more readily understood) name such as `cloud-console`. I'll wait til Breda gets back and maybe we can settle on names that we agree on before I add them upstream. WDYT?

smccarthy-ie:
@jbyrne-redhat, +1, sounds good, `cloud-console` should also help guard against next rebrand :)

jbyrne-redhat:
> @jbyrne-redhat, +1, sounds good, `cloud-console` should also help guard against next rebrand :)

Exactly, yes!

PR #565, opened 2022-08-05
Docs: Minor style updates

jbyrne-redhat:
From conversation with @riccardo-forina, I need to merge something upstream to trigger a rebuild of the stage UI. Applying these minor style changes for that purpose.


riccardo-forina (approved):
:shipit:

PR #568, opened 2022-08-15
docs: added connector guide for rhoas cli

jackdelahunt:


wtrocki:
CC @MelissaFlinn We have decided to help connectors team by providing the same style of guide as we have for registry and kafka. This is good start for us to extend specific guides as required.

MelissaFlinn:
OK - thanks for letting me know.  I will review it.

wtrocki (approved):

wtrocki:
This should have -cli prefix as all other guides. 
Entire guide assumes that user has namespace. We need to actually create it (we cannot assume it exist)

MelissaFlinn:
I  believe that this README.adoc file belongs under a new subdirctory with this path: app-services-guides/docs/connectors/ rhoas-cli-getting-started-connectors (rather than under the `rhoas` subdirectory)

to follow the pattern of the existing cli guides:
app-services-guides/docs/registry/rhoas-cli-getting-started-registry
and 
pp-services-guides/docs/kafka/rhoas-cli-getting-started-kafka

jackdelahunt:
> Entire guide assumes that user has namespace. We need to actually create it (we cannot assume it exist)

Before connector creation I say in the guide to create a new namespace, is this what you mean? Not sure.

PR #572, opened 2022-08-19
chore: don't eagerly load dependencies

riccardo-forina:
remove all eager loading as we should never eagerly load any dependency ever; when eagerly loading, the dependency gets bundled with the bundle that will be loaded through module federation, forcing users to download them all over again


wtrocki (commented):
LGTM :)

PR #577, opened 2022-08-26
jc-1084  update the connectors cli get started doc

MelissaFlinn:
updated the rhoas connector CLI guide


MelissaFlinn:
there is an error with this PR

PR #585, opened 2022-09-15
docs: Update external documentation links on Learning Resources page

jbyrne-redhat:
Verified changes using a local build, as described [here](https://github.com/redhat-developer/app-services-guides/blob/main/CONTRIBUTING.adoc#previewing-the-quick-starts-locally).


.build/external-documentation/sr-core-api-reference.quickstart.yml
```diff
@@ -3,16 +3,16 @@ kind: QuickStarts
...
   annotations:
-    order: 7
+    order: 9
 spec:
   version: 0.1
   type:
     text: Documentation
     color: orange
-  displayName: Service Registry core REST API Reference
+  displayName: OpenShift Service Registry instance API reference
```
bredamc, line 12:
```suggestion
  displayName: OpenShift Service Registry core instance API reference
```

.build/external-documentation/sr-core-api-reference.quickstart.yml
```diff
@@ -3,16 +3,16 @@ kind: QuickStarts
...
-  description: Learn how to use the Apicurio Registry core REST API to manage the contents of a Service Registry instance.
+  description: Learn how to use the OpenShift Service Registry instance API to manage the contents of a Service Registry instance.
```
bredamc, line 15:
```suggestion
  description: Learn how to use the OpenShift Service Registry core instance API to manage the contents of a Service Registry instance.
```

bredamc (commented):
LGTM, John -- many thanks. Just two minor comments about including "core" in the Service Registry instance entries, for consistency with the text in the UI.

.build/external-documentation/sr-core-api-reference.quickstart.yml
bredamc:
```suggestion
  description: Learn how to use the OpenShift Service Registry instance API (also known as the _Core Registry API_) to manage the contents of a Service Registry instance. 
```

bredamc:
I think that's a fair assessment, John -- I've started a discussion elsewhere about the inconsistency between the UI and the API doc, but we should refer to the actual title as it is now. Perhaps we could update the description (as suggested in my additional comment above) to clarify the correlation?

Also, do the files in the `.build` directory use the attributes specified in the `docs/_artifacts` directory?

.build/external-documentation/sr-core-api-reference.quickstart.yml
jbyrne-redhat: It's a good suggestion, but the description is limited to just 3 lines. Adding that additional text will give us a truncated description, as shown in the figure:

image

jbyrne-redhat:
> I think that's a fair assessment, John -- I've started a discussion elsewhere about the inconsistency between the UI and the API doc, but we should refer to the actual title as it is now. Perhaps we could update the description (as suggested in my additional comment above) to clarify the correlation?
> 
> Also, do the files in the `.build` directory use the attributes specified in the `docs/_artifacts` directory?

Thanks, Breda. I replied to your comment. I think we might have to accept the inconsistency that exists between the API and UI for now.

The files in the `.build` directory don't reference the doc attributes. Whether they _could_, is maybe a conversation for another day. For now, we need to take care with our references to product names, etc.

bredamc:
Ah ok! Let's go with the original text then so, John -- sorry for all the messing around. I'll approve now :)

bredamc (approved):
LGTM -- thanks, John!

PR #592, opened 2022-10-12
Fix conditional for verification step

bhardesty:


bhardesty:
@jbyrne-redhat When publishing the docs, I noticed that this quick start was missing a conditional on the final verification, which caused the quick start verification to be included incorrectly in the published version.

jbyrne-redhat:
> @jbyrne-redhat When publishing the docs, I noticed that this quick start was missing a conditional on the final verification, which caused the quick start verification to be included incorrectly in the published version.

@bhardesty Thanks for moving to fix this. I'm not actually sure we have anything to do, however. In that final task, you're working on the command line. So, regardless of whether you're following the quick start in the console or from the Customer Portal, I think the same wording for the verification step applies in each case. WDYT?

bhardesty:
@jbyrne-redhat Yes, I think that's correct. However, the verification here is in the form of a question. I think we've only been using that form for the quick starts, not the customer portal. But when I tried to restate the verification in a different form that would be appropriate for the customer portal, it became so obvious that it seemed unnecessary to even include. So that's how I ended up conditionalizing the verification for the quick start only, and not having a verification at all for the customer portal.

jbyrne-redhat (approved):

PR #596, opened 2022-10-20
Amended Tech Preview text in line with CCS guidelines.

AedinC:
Preview available [here](http://file.ams.redhat.com/aecollin/JK-9956/README.html).

### DIFF
--- a/docs/kafka/getting-started-kafka/README.adoc
+++ b/docs/kafka/getting-started-kafka/README.adoc
@@ -137,9 +137,11 @@ NOTE: The cloud provider is preconfigured.
 * *Cloud region*: Select `US East, N. Virginia`.
 * *Availability zones*: Select `Multi`.
 ////
-. If you've purchased a subscription for {product-kafka}, use the slider to select an instance size. Instances with more streaming units have more capacity.
+. If you've purchased a subscription for {product-kafka}, use the slider to select an instance size. Instances with more streaming units have more capacity. If you're creating a trial instance, the instance size is preconfigured.
 +
-NOTE: In {product-kafka}, you can create Kafka instances of 1 or 2 streaming units. Instances larger than 1 streaming unit are currently available only as a Technology Preview and should not be used in production. If you're creating a trial instance, the instance size is preconfigured.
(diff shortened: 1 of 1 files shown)


docs/kafka/getting-started-kafka/README.adoc
```diff
@@ -137,9 +137,11 @@ NOTE: The cloud provider is preconfigured.
...
 +
-NOTE: In {product-kafka}, you can create Kafka instances of 1 or 2 streaming units. Instances larger than 1 streaming unit are currently available only as a Technology Preview and should not be used in production. If you're creating a trial instance, the instance size is preconfigured.
+NOTE: In {product-kafka}, you can create Kafka instances of 1 or 2 streaming units. Instances larger than 1 streaming unit are currently available as a Technology Preview feature only. Technology Preview features are not supported with Red Hat production service level agreements (SLAs) and might not be functionally complete. Red Hat does not recommend using them in production. These features provide early access to upcoming product features, enabling
+customers to test functionality and provide feedback during the development process.
+For more information about the support scope of Red Hat Technology Preview features, see https://access.redhat.com/support/offerings/techpreview/.
```
jbyrne-redhat, line 144: Hey, @AedinC. Great idea to update this text.

I know this the boilerplate text that we discussed in the CCS Best Practices meeting. But I also think we have a _bit_ of leeway to make a slight tweak in our implementation of it. I feel quite strongly that "Red Hat does not recommend using them in production" should be the final thing we say. Right now, this critical statement gets slightly buried.

I would reformulate the note slightly as:

```

NOTE: In {product-kafka}, you can create Kafka instances of 1 or 2 streaming units. Instances larger than 1 streaming unit are currently available as a Technology Preview feature only. Technology Preview features provide early access to upcoming product features, enabling customers to test functionality and provide feedback during the development process. Technology Preview features are not supported with Red Hat production service level agreements (SLAs) and might not be functionally complete. Red Hat does not recommend using them in production. 

```
WDYT? I don't think we're straying off piste in any dramatic way. I'm happy to follow up with Nacia to give this same feedback.

AedinC: Thanks for the feedback @jbyrne-redhat. Happy to tweak it as you suggest but still think it would be good to include the link for extra info.

jbyrne-redhat: Yes, sorry - I didn't mean to drop that final sentence with the link! :roll_eyes:

jbyrne-redhat (approved):

docs/kafka/getting-started-kafka/README.adoc
AedinC: Cool, thanks!

PR #603, opened 2022-11-11
Updated access permissions intro in Managing Access guide.

AedinC:
Updated intro to guide as per recommendations from [MGDSTRM-9861](https://issues.redhat.com/browse/MGDSTRM-9861).
Note to reviewers: preview available [here](http://file.ams.redhat.com/aecollin/JK-9861/README.html).

### DIFF
--- a/docs/kafka/access-mgmt-kafka/README.adoc
+++ b/docs/kafka/access-mgmt-kafka/README.adoc
@@ -89,7 +89,9 @@ ifdef::context[:parent-context: {context}]
 
 // Purpose statement for the assembly
 [role="_abstract"]
-As an owner of a Kafka instance in {product-long-kafka}, you can manage the level of access that other user accounts and service accounts have to your instance. You can allow or deny access to your instance for specific accounts or for all accounts in your organization. You can also allow other users or service accounts to manage the level of access to your instance for you.
+As an owner of a Kafka instance in {product-long-kafka}, you control immutable access permissions on all resources within that Kafka instance including topics, groups, transactional IDs, and Access Control Lists (ACLs). These permissions are implicit meaning that other users do not see them listed in the {product-long-kafka} web console or from Kafka APIs.
+
+You can use these permissions to manage the level of access that other user accounts and service accounts have to your instance. You can allow or deny access to your instance for specific accounts or for all accounts in your organization. You can also allow other users or service accounts to manage the level of access to your instance for you.
 
 You can manage access for only the Kafka instances that you create or for instances that the owner has enabled you to access and alter.
 
--- a/docs/kafka/access-mgmt-kafka/README.adoc
+++ b/docs/kafka/access-mgmt-kafka/README.adoc
@@ -93,7 +93,7 @@ As an owner of a Kafka instance in {product-long-kafka}, you can manage the leve
 
 You can manage access for only the Kafka instances that you create or for instances that the owner has enabled you to access and alter.
 
-You also have a set of permissions for all resources within a Kafka instance including topics, groups, transactional IDs, and Access Control Lists (ACLs). This set of permissions cannot be changed and cannot be seen by you or any other user in the *Permission* list in the {product-long-kafka} web console or from Kafka APIs.
+As an instance owner, you automatically have a set of permissions for all resources within a Kafka instance including topics, groups, transactional IDs, and Access Control Lists (ACLs). This set of permissions cannot be changed and cannot be seen by you or any other user in the *Permission* list in the {product-long-kafka} web console or from Kafka APIs.
 
 //Additional line break to resolve mod docs generation error, not sure why. Leaving for now. (Stetson, 20 May 2021)
 
--- a/docs/kafka/access-mgmt-kafka/README.adoc
+++ b/docs/kafka/access-mgmt-kafka/README.adoc
@@ -89,11 +89,11 @@ ifdef::context[:parent-context: {context}]
 
 // Purpose statement for the assembly
 [role="_abstract"]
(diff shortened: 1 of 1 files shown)


docs/kafka/access-mgmt-kafka/README.adoc
```diff
@@ -85,7 +85,9 @@ ifdef::context[:parent-context: {context}]
 
 // Purpose statement for the assembly
 [role="_abstract"]
-As an owner of a Kafka instance in {product-long-kafka}, you can manage the level of access that other user accounts and service accounts have to your instance. You can allow or deny access to your instance for specific accounts or for all accounts in your organization. You can also allow other users or service accounts to manage the level of access to your instance for you.
+As an owner of a Kafka instance in {product-long-kafka}, you control immutable access permissions on all resources within that Kafka instance including topics, groups, transactional IDs, and Access Control Lists (ACLs). These permissions are implicit meaning that other users do not see them listed in the {product-long-kafka} web console or from Kafka APIs.
```
jbyrne-redhat, line 88: A couple of comments:

- _Immutable_ means "cannot be changed". I _think_ Keith meant that the level of control an owner has by default is immutable, not that access permissions are immutable. Access permissions _can_ be changed, for example, to grant access to other accounts.
- I don't think we need to introduce a term such as _implicit,_ which then requires definition. We can probably say this in plainer language. 

Therefore, I suggest a rewrite like the following:

```
As an owner of a Kafka instance in {product-long-kafka}, you control access permissions for all resources within that Kafka instance by default. This includes topics, groups, transactional IDs, and Access Control Lists (ACLs) themselves. Other users cannot see permissions for your Kafka instance in the {product-long-kafka} web console or from Kafka APIs.
```

docs/kafka/access-mgmt-kafka/README.adoc
```diff
@@ -85,7 +85,9 @@ ifdef::context[:parent-context: {context}]
...
+
+You can use these permissions to manage the level of access that other user accounts and service accounts have to your instance. You can allow or deny access to your instance for specific accounts or for all accounts in your organization. You can also allow other users or service accounts to manage the level of access to your instance for you.
```
jbyrne-redhat, line 90: We don't really introduce the concept of ACL permissions and how to modify them until the next section. So maybe avoid the phrase "these permissions" here and let the next section do the talking.

Therefore:

"You can manage the level of access that other user accounts and service accounts have to your instance..."

docs/kafka/access-mgmt-kafka/README.adoc
```diff
@@ -85,7 +85,9 @@ ifdef::context[:parent-context: {context}]
 
 // Purpose statement for the assembly
 [role="_abstract"]
-As an owner of a Kafka instance in {product-long-kafka}, you can manage the level of access that other user accounts and service accounts have to your instance. You can allow or deny access to your instance for specific accounts or for all accounts in your organization. You can also allow other users or service accounts to manage the level of access to your instance for you.
+As an owner of a Kafka instance in {product-long-kafka}, you control access permissions for all resources within that Kafka instance by default. This includes topics, groups, transactional IDs, and Access Control Lists (ACLs) themselves. Other users cannot see permissions for your Kafka instance in the {product-long-kafka} web console or from Kafka APIs.
```
k-wall, line 88: The owner's permissions are "special" - they are inherent (cannot be changed) and they are don't appear in the permission list for either the owner or other users.

* "by default" - to me conveys an initial state and leaves open the possibility that things can be changed.  That's not the case here.  The owner's permissions are set in stone.
* "Other users cannot see permissions for your Kafka instance in the {product-long-kafka} web console or from Kafka APIs." is unclear.     It is not just "other users" who don't see the owner's permission.  It is the owner too.   Also we are not talking about all permissions - just this special set of owner permisisons.




This is not conveying that 

this set of permissions held by the owner is immutable.
this set of permissions is implicit - that means to say that user won't see these permissions enumerated in the console or via the API.

AedinC: Thanks for the clarification @k-wall . Would something like this work (trying to keep the language as plain as possible for better readability):

You also control a set of permissions for all resources within a Kafka instance including topics, groups, transactional IDs, and Access Control Lists (ACLs). This set of permissions cannot be changed and cannot be seen by you or any other user in the **Permission** list in the {product-long-kafka} web console or from Kafka APIs.

docs/kafka/access-mgmt-kafka/README.adoc
```diff
@@ -93,6 +93,8 @@ As an owner of a Kafka instance in {product-long-kafka}, you can manage the leve
 
 You can manage access for only the Kafka instances that you create or for instances that the owner has enabled you to access and alter.
 
+You also have a set of permissions for all resources within a Kafka instance including topics, groups, transactional IDs, and Access Control Lists (ACLs). This set of permissions cannot be changed and cannot be seen by you or any other user in the *Permission* list in the {product-long-kafka} web console or from Kafka APIs.
```
jbyrne-redhat, line 96: Looks good to me, @AedinC. Suggesting a couple of final, very minor tweaks:

"*As an instance owner*, you *automatically have full access permissions* for all resources within the instance, including...."

jbyrne-redhat (approved):
LGTM. 

Suggested two minor tweaks that might help underscore a couple of important points.

docs/kafka/access-mgmt-kafka/README.adoc
AedinC: Thanks John. :)

k-wall (approved):
lgtm

PR #607, opened 2022-11-25
docs: JK-6383: update guides to correctly describe configured ACL permissions

jbyrne-redhat:
**Notes for reviewers**

- As discussed in https://issues.redhat.com/browse/MGDSTRM-6383, removing incorrect statement (in a few guides) that says that the example ACL permissions shown allow applications to "use any producer". This is not the case.
- Also as discussed in the JIRA, there are a couple of instances where we incorrectly state that the example ACL permissions allow applications associated with the service account to delete topics. In fact, only the "Create" operation is set.
- HTML previews for the _Getting Started_ and _Access Management_ guides are [here](http://file.emea.redhat.com/~jbyrne/JK-6383/Getting%20started%20with%20OpenShift%20Streams%20for%20Apache%20Kafka.html) and [here](http://file.emea.redhat.com/~jbyrne/JK-6383/Managing%20account%20access%20in%20OpenShift%20Streams%20for%20Apache%20Kafka.html).

### DIFF
--- a/docs/kafka/access-mgmt-kafka/README.adoc
+++ b/docs/kafka/access-mgmt-kafka/README.adoc
@@ -200,7 +200,7 @@ a|
 * `Describe`
 |===
 
-By default, new Kafka instances contain the permissions shown in the following table. These permissions allow all accounts in the organization to view the instance permissions and to view topics in the instance, but not to produce or consume messages.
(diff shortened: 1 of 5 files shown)


katheris (approved):
The changes look good to me.

PR #608, opened 2022-11-29
docs:JK-10189: Removed info about tech preview.

AedinC:


fmcdonal (commented):
In {product-kafka}, you can create Kafka instances of 1 or 2 streaming units depending on the capacity required.

fmcdonal (commented):
I think this line is misleading because it feels like you can have more than 2 streaming units, the word "more" indicates more than what is stated in the last line so I suggest you integrate both lines to:
"In {product-kafka}, you can create Kafka instances of 1 or 2 streaming units depending on the capacity required."

PR #609, opened 2022-12-09
chore: update the setup ruby action to the newer one

riccardo-forina:


bredamc (approved):
Not sure if my approval is valid (given that I'm not familiar with the code) but adding my approval because I'm waiting for this fix so that I can test my own PR :)

PR #624, opened 2023-01-19
Add "scope" to sasl.jaas.config

grdryn:
When using the OAuthBearerLoginCallbackHandler (as specified as the value for `sasl.login.callback.handler.class` in this example config), and when using `sso.redhat.com` (or potentially other Keycloak and maybe some other OAuth servers), which we do, then without specifying a scope will currently lead to a response with an empty scope, which is invalid and leads to failure.

The `sso.redhat.com` folks have suggested `openid` as the scope value, and I've verified that it works.

This seems to be the only place where `org.apache.kafka.common.security.oauthbearer.secured.OAuthBearerLoginCallbackHandler` is used in this repo, I'm not sure if it's used in any other docs or quickstarts?

Where `io.strimzi.kafka.oauth.client.JaasClientOauthLoginCallbackHandler` is used in some other cases here instead, then scope does not need to be set.

Whether we should either change all cases to `org.apache.kafka.common.security.oauthbearer.secured.OAuthBearerLoginCallbackHandler`, or otherwise specify a scope in both cases, is not a question I have an answer to.

### DIFF
--- a/docs/kafka/kafka-bin-scripts-kafka/README.adoc
+++ b/docs/kafka/kafka-bin-scripts-kafka/README.adoc
@@ -152,8 +152,9 @@ security.protocol=SASL_SSL
 sasl.oauthbearer.token.endpoint.url= {sso-token-url}
 
 sasl.jaas.config=org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required \
-clientId="<client-id>" \
-clientSecret="<client-secret>" ;
+  scope="openid" \
+  clientId="<client-id>" \
+  clientSecret="<client-secret>" ;
 
 sasl.login.callback.handler.class=org.apache.kafka.common.security.oauthbearer.secured.OAuthBearerLoginCallbackHandler
 ----


jbyrne-redhat (approved):
I'm not sure if you're awaiting any other eyes on this, @grdryn, but from a docs perspective, this looks good to me.

Like you, this is the only occurrence of `org.apache.kafka.common.security.oauthbearer.secured.OAuthBearerLoginCallbackHandler` that I can see in the `app-service-guides` repo.

The example Quarkus applications that we use in our RHOSAK and RHOSR quick starts use the Strimzi OAuth handler: https://github.com/redhat-developer/app-services-guides/blob/main/code-examples/quarkus-kafka-quickstart/src/main/resources/application.properties#L32.

The only other question I have is about the [example Node.js application](https://github.com/nodeshift-starters/reactive-example) that we use in one of our quick starts. I can't tell what that uses, but right now we [don't instruct users](https://access.redhat.com/documentation/en-us/red_hat_openshift_streams_for_apache_kafka/1/guide/5b6badee-eaf3-4a10-b5ec-57dc6b94ec0f#_047f969b-3473-4752-8738-8ed5f55f7a0f) to set a scope. Do you know what that particular application uses?

grdryn:
> I'm not sure if you're awaiting any other eyes on this, @grdryn, but from a docs perspective, this looks good to me.

I'm not, so if you have the ability to merge, please do, thanks! :slightly_smiling_face: 

> The only other question I have is about the [example Node.js application](https://github.com/nodeshift-starters/reactive-example) that we use in one of our quick starts. I can't tell what that uses, but right now we [don't instruct users](https://access.redhat.com/documentation/en-us/red_hat_openshift_streams_for_apache_kafka/1/guide/5b6badee-eaf3-4a10-b5ec-57dc6b94ec0f#_047f969b-3473-4752-8738-8ed5f55f7a0f) to set a scope. Do you know what that particular application uses?

The problem that this PR addresses is specific to something in Kafka's Java client library, so it won't affect that Javascript example. Looking at that project, it appears to be using kafkajs: https://github.com/nodeshift-starters/reactive-example/blob/kafkajs/producer-backend/package.json#L20

If that works without a scope for now, then that's fine, just like the Java examples that use the Strimzi handler. There might come a time in the future when `sso.redhat.com` will enforce a scope or something, and at that point we'd have to go back and look at everything anyway. :)

jbyrne-redhat:
> The problem that this PR addresses is specific to something in Kafka's Java client library, so it won't affect that Javascript example. Looking at that project, it appears to be using kafkajs: https://github.com/nodeshift-starters/reactive-example/blob/kafkajs/producer-backend/package.json#L20
>
Ah, OK, that makes sense. Thanks for the additional explanation, Ger.

Will merge now.

PR #631, opened 2023-02-02
docs: JX-326: Removes redundant 'Learn how to' text

bredamc:


jbyrne-redhat (approved):

docs/kafka/kcat-kafka/README.adoc
```diff
@@ -126,7 +126,7 @@ Version 1.7.0 (librdkafka 1.3.0 builtin.features=gzip,snappy,ssl,sasl,regex,lz4,
 ifdef::qs[]
 [#description]
 ====
-Learn how to use Kcat to interact with a Kafka instance in {product-long-kafka}.
+Use Kcat to interact with a Kafka instance in {product-long-kafka}.
```
smccarthy-ie, line 129: Should this be `Kcat` or `kcat`? All lowercase in the GitHub project: https://github.com/edenhill/kcat

jbyrne-redhat: That's a really good point, @smccarthy-ie  - I don't know why this hasn't jumped out to us before now :roll_eyes: :laughing: 

I don't see a single instance in the GH project (nor the Confluent documentation, for that matter) of it being called `Kcat`. If there was an original discussion to take some licence and capitalize `Kafkacat` when we first documented it, I'm not aware of it.

I propose leaving it unchanged in this PR and I will create a dedicated task to track down all instances and update them. WDYT?

smccarthy-ie: Yes, sounds good, shouldn't block this PR. 

Will need to be changed in a few places anyway. Thanks!

smccarthy-ie (approved):

PR #637, opened 2023-02-07
fix(helm-example): use rhosak org

rkpattnaik780:


jackdelahunt (approved):

jbyrne-redhat (approved):
Looks good to me, @rkpattnaik780. No documentation impact arising from this change. We make reference to the `helm-kafka-example` code example in the doc, but not the location of the underlying image.

PR #638, opened 2023-02-08
feat: add example for java client

rkpattnaik780:
The example added here is related to the code snippets displayed in console. 
Having the entire sample application here will help in future if we want to show entire code example and they will also be used in e2e testing.

### DIFF
--- a/code-examples/kafka-java-maven/src/main/java/org/example/ProducerExample.java
+++ b/code-examples/kafka-java-maven/src/main/java/org/example/ProducerExample.java
@@ -1,0 +1,23 @@
+package org.example;
+
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.common.serialization.StringSerializer;
+
+public class ProducerExample {
+
+    public static void main(String[] args) {
+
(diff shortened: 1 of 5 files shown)


code-examples/kafka-java-maven/src/main/java/org/example/ProducerExample.java
```diff
@@ -0,0 +1,23 @@
...
+public class ProducerExample {
+
+    public static void main(String[] args) {
+
+        //Creating producer properties
+        var properties= KafkaConfig.properties();
+        properties.setProperty(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName());
+        properties.setProperty(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, StringSerializer.class.getName());
+
+        KafkaProducer<String,String> producer= new KafkaProducer<String,String>(properties);
```
jackdelahunt, line 17: We already use `var` in this example so I think reducing clutter here on line 17 will be good by using it there aswell

rkpattnaik780: You mean something like `var producer = ` ?

jackdelahunt: Yes exactly

PR #661, opened 2023-04-03
Deleting the service binding kafka card .yml from the .build directory

fmcdonal:
Needs to be removed from the console as ENTERPRISE does not support RJOAS operator for service binding


smccarthy-ie:
LGTM, thanks!

smccarthy-ie (approved):

PR #676, opened 2024-01-15
Update README.adoc with alternative products

smccarthy-ie:
Add note to clarify that these cloud services are no longer available and point at self-managed alternative products instead.


MelissaFlinn:
lgtm

MelissaFlinn (approved):
lgtm

